PR #10, opened 2023-01-17
#7 Add dynamic navigation based on screen size

### LINKED ISSUE #7: Add Navigation Rail
Use Material design as per: https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#navigationrail
Use Navigation Component: 

Tabs: Home, Camera, Profile - empty views

michal-kucznerowicz:
Add dynamic navigation depending on the current screen size based on Material 3 design guidelines. Of course, we can change the breakpoints of each type to suit our needs.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -52,12 +52,15 @@ dependencies {
     implementation composeBom
     androidTestImplementation composeBom
 
-    implementation 'androidx.core:core-ktx:1.9.0'
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.5.1'
-    implementation 'androidx.activity:activity-compose:1.6.1'
     implementation 'androidx.compose.ui:ui'
     implementation 'androidx.compose.ui:ui-tooling-preview'
     implementation 'androidx.compose.material3:material3'
+    implementation 'androidx.compose.material:material-icons-extended'
+    implementation 'androidx.compose.material3:material3-window-size-class'
+    implementation 'androidx.core:core-ktx:1.9.0'
+    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.5.1'
+    implementation 'androidx.activity:activity-compose:1.6.1'
+    implementation 'androidx.navigation:navigation-compose:2.5.3'
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
--- a/app/src/main/java/com/example/android/picover/MainActivity.kt
+++ b/app/src/main/java/com/example/android/picover/MainActivity.kt
@@ -3,38 +3,75 @@ package com.example.android.picover
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
-import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.material3.MaterialTheme
-import androidx.compose.material3.Surface
-import androidx.compose.material3.Text
-import androidx.compose.runtime.Composable
+import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
+import androidx.compose.material3.windowsizeclass.calculateWindowSizeClass
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.tooling.preview.Preview
-import com.example.android.picover.ui.theme.PicoverTheme
+import androidx.navigation.compose.rememberNavController
+import com.example.android.picover.presentation.navigation.NavigationItem
+import com.example.android.picover.presentation.navigation.NavigationType
+import com.example.android.picover.presentation.navigation.PicoverNavigationBar
+import com.example.android.picover.presentation.navigation.PicoverNavigationDrawer
+import com.example.android.picover.presentation.navigation.PicoverNavigationRail
+import com.example.android.picover.presentation.ui.theme.PicoverTheme
 
+@OptIn(ExperimentalMaterial3WindowSizeClassApi::class, ExperimentalMaterial3Api::class)
 class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
             PicoverTheme {
-                // A surface container using the 'background' color from the theme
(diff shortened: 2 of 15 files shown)


BorysChajdas:
Would be better to separate the commits into two pull requests 😄 Changes regarding the .gitignore are not connected with this feature, so will be better to review them separately.

app/build.gradle
```diff
@@ -52,12 +52,16 @@ dependencies {
...
 
-    implementation 'androidx.core:core-ktx:1.9.0'
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.5.1'
-    implementation 'androidx.activity:activity-compose:1.6.1'
     implementation 'androidx.compose.ui:ui'
     implementation 'androidx.compose.ui:ui-tooling-preview'
     implementation 'androidx.compose.material3:material3'
+    implementation 'androidx.compose.material:material-icons-extended'
+    implementation 'androidx.compose.material3:material3-window-size-class'
+
```
BorysChajdas: Why new line :D?

app/src/main/java/com/example/android/picover/MainActivity.kt
```diff
@@ -1,40 +0,0 @@
-package com.example.android.picover
-
-import android.os.Bundle
-import androidx.activity.ComponentActivity
```
BorysChajdas, line 4: It will be better to move changing the package to another commit (and PR) to keep the history of `MainActivity`.

app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,6 @@
 <resources>
     <string name="app_name">Picover</string>
+    <string name="home">Home</string>
```
BorysChajdas, line 3: Just to inform: I've added issue to discuss the naming convention for the strings: https://github.com/intive/Picover/issues/13

app/src/main/java/com/example/android/picover/presentation/navigation/NavigationType.kt
```diff
@@ -0,0 +1,19 @@
...
+import androidx.compose.material3.windowsizeclass.WindowWidthSizeClass
+
+enum class NavigationType {
+    NAVIGATION_BAR,
+    NAVIGATION_RAIL,
+    NAVIGATION_DRAWER;
+
+    companion object {
+        fun fromWindowSize(windowSize: WindowSizeClass) = when (windowSize.widthSizeClass) {
+            WindowWidthSizeClass.Compact -> NAVIGATION_BAR
```
BorysChajdas: Redundant branch because `NAVIGATION_BAR` is in `else` block.

BorysChajdas (changes requested):

app/build.gradle
michal-kucznerowicz: To visually clarify which dependencies are automatically updated.

app/src/main/java/com/example/android/picover/MainActivity.kt
michal-kucznerowicz: Done

app/src/main/java/com/example/android/picover/presentation/navigation/NavigationType.kt
michal-kucznerowicz: Done

michal-kucznerowicz:
> Would be better to separate the commits into two pull requests 😄 Changes regarding the .gitignore are not connected with this feature, so will be better to review them separately.

[https://github.com/intive/Picover/pull/15](url)

app/build.gradle
marcinWisnia: is there any benefit of knowledge that the dependencies above the empty line are automatically updated and the rest not :D ?

app/src/main/java/com/example/android/picover/MainActivity.kt
marcinWisnia: shouldn't @BorysChajdas resolve this conversation after checking if it was excluded to a separate MR :D ?

michal-kucznerowicz: Yes, but I don't remember how reviews work on GitHub 😄 that's why I resolved it. Additionally, we discussed offline that it's ok for now and we will work out some conventions at Thursday's meeting.

app/build.gradle
michal-kucznerowicz: It seems not 😄. I can remove it.

michal-kucznerowicz: Done

app/build.gradle
```diff
@@ -52,12 +52,15 @@ dependencies {
     implementation composeBom
     androidTestImplementation composeBom
 
-    implementation 'androidx.core:core-ktx:1.9.0'
-    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.5.1'
-    implementation 'androidx.activity:activity-compose:1.6.1'
     implementation 'androidx.compose.ui:ui'
     implementation 'androidx.compose.ui:ui-tooling-preview'
     implementation 'androidx.compose.material3:material3'
+    implementation 'androidx.compose.material:material-icons-extended'
```
BorysChajdas, line 58: If I remember correctly it is not recommended to add it because it is very heavy in terms of size :D It increases APK size a lot, but probably we can think about it later.

app/src/main/java/com/example/android/picover/presentation/MainActivity.kt
```diff
@@ -0,0 +1,77 @@
+package com.example.android.picover.presentation
```
BorysChajdas: There is still a problem with `MainActivity` because it is shown as deleted and added a new file :D. As I said in another comment – probably the change of package should be done in another commit (and another pull request).

BorysChajdas (changes requested):

app/src/main/java/com/example/android/picover/MainActivity.kt
BorysChajdas: I think it is still not resolved :D because MainActivity is showing as deleted

app/src/main/java/com/example/android/picover/presentation/MainActivity.kt
michal-kucznerowicz: I think now it's done. Also, I changed PR to have only one commit and I will create a new PR with moving `MainActivity`.

app/build.gradle
michal-kucznerowicz: You're right. It is said at the maven repository https://mvnrepository.com/artifact/androidx.compose.material/material-icons-extended. If we're only going to use a few of them I think a better way would be to add drawables to our resources for only those icons that we need.

BorysChajdas (approved):

marcinWisnia (approved):

bartek977 (approved):

BorysChajdas:
@michal-kucznerowicz I will merge it as it has 3 approves 😄

PR #15, opened 2023-01-17
#12 Update .gitignore and remove files that should not be included in vcs

### LINKED ISSUE #12: Improve .gitignore

michal-kucznerowicz:


BorysChajdas (approved):

kmaslowiec:
@michal-kucznerowicz can you please add also this file?
image

marcinWisnia (approved):

bartek977 (approved):

michal-kucznerowicz:
> @michal-kucznerowicz can you please add also this file? image

Done

Nataniel-Antosik:
Could you rebase changes to this PR to add CI

.gitignore
```diff
@@ -8,6 +8,12 @@
 /.idea/workspace.xml
 /.idea/navEditor.xml
 /.idea/assetWizardSettings.xml
+/.idea/misc.xml
+/.idea/compiler.xml
+/.idea/gradle.xml
+/.idea/kotlinc.xml
+/.idea/vcs.xml
+/.idea/.name
```
MarekMacko: Which files do we really need from `/idea` directory? Seems that we can exclude whole directory instead of listing each file/subdirectory.

.gitignore
```diff
@@ -25,5 +31,4 @@ captures/
 lint/intermediates/
 lint/generated/
 lint/outputs/
-lint/tmp/
-
+lint/tmp/
```
MarekMacko: Probably it's redundant as well, since it's a "top-level" `.gitignore` file. Lint files are generated under `<module>/build/reports` and it's currently excluded by `app/.gitignore` file, note `/build` inside.

.gitignore
michal-kucznerowicz: It looks like we don't need anything from `/.idea`. Only if we want to add copyrights we can exclude this directory later.

.gitignore
michal-kucznerowicz: Seems like you're right, so I will remove it.

.gitignore
michal-kucznerowicz: Done

.gitignore
michal-kucznerowicz: Done

.gitignore
```diff
@@ -21,9 +15,4 @@ local.properties
 bin/
 gen/
 out/
-captures/
-lint/intermediates/
-lint/generated/
-lint/outputs/
-lint/tmp/
-
+captures/
```
MarekMacko: Please check if rest of the directories can be removed as well

michal-kucznerowicz: Done

PR #16, opened 2023-01-17
#4 setup CI for pull request and 'dev' branch

### LINKED ISSUE #4: Setup CI
- [ ] build debug

kmaslowiec:


.github/workflows/android.yml
```diff
@@ -0,0 +1,26 @@
+name: Android CI
```
Nataniel-Antosik, line 1: Minor: You could make the name more meaningful like a _Verification pull request_ or something else

.github/workflows/android.yml
```diff
@@ -0,0 +1,26 @@
...
+      uses: actions/setup-java@v3
+      with:
+        java-version: '11'
+        distribution: 'temurin'
+        cache: gradle
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+    - name: Build with Gradle
+      run: ./gradlew build
```
Nataniel-Antosik, line 26: You could consider if we needed build to the verification pull request.

Nataniel-Antosik:
Also, you could make two different yaml files for CI one for the pull requests and one for develop branch after the merge.

.github/workflows/android.yml
Nataniel-Antosik: You could make the name more meaningful like a _Verification pull request_ or something else

Nataniel-Antosik (changes requested):

.github/workflows/android.yml
Nataniel-Antosik: You could consider if we needed build to the verification pull request.

Nataniel-Antosik (changes requested):

.github/workflows/android.yml
kmaslowiec: Currently there is only one CI flow. It can be changed if necessary

.github/workflows/android.yml
kmaslowiec: The build is verified every time a new Pull Request is created and once it is pushed to the default branch. It is better to be sure that branch in PR works before we can check it.

Nataniel-Antosik: Ok, but I think about `./gradlew assemble` for the pull request and `./gradlew build` for develop

.github/workflows/android.yml
Nataniel-Antosik: Yes, I see but in the actions filter, we could see only `Android CI`
image
 instead of for example: 
image
We can have this if we have two flows one for the dev branch and one for the pull request. If it is not important for us you can leave it as is now :D

Nataniel-Antosik (approved):

.github/workflows/android.yml
kmaslowiec: Actually update the name is a good point, I will take care of it tomorrow

.github/workflows/android.yml
BorysChajdas: We should make sure that unit tests are checked during the verification of PR, but if I'm not wrong `build` executes the unit tests

BorysChajdas:
The ".idea" files should not be included here

.github/workflows/android.yml
Nataniel-Antosik: Yes command **build** executes the _tests_ so if we want to use _assemble_ for pull requests also we need to check another job for _tests_

kmaslowiec: The Build definitely includes the tests:
image
@Nataniel-Antosik Is it possible to provide hard data on how beneficial creating assemble job instead of one build would be? Is it a common practice or just work around in case of significantly big projects?

kmaslowiec:
> 



> The ".idea" files should not be included here
@BorysChajdas files excluded as suggested

kmaslowiec:
> Also, you could make two different yaml files for CI one for the pull requests and one for develop branch after the merge.

@Nataniel-Antosik I have divided the workflows as you suggested. Later on it can be necessary to make different jobs for pull requests and dev branch.

.github/workflows/android.yml
Nataniel-Antosik: If we decide we don't want everything that offers **build** command we could split this commend on some jobs for the pull requests. I tested the command **build** and some jobs executed simultaneously doing the almost same work like **build** and splitting command **build** is faster. What about the common practice I believe it depends on the project.

bartek977 (approved):

Issue #24, opened 2023-01-26
Take a picture


BorysChajdas:
On Camera screen:
- Launch the device camera to take the picture,
- Show the picture after taking it.

MarekMacko:
@BorysChajdas Is there something to do or can we close it?

PR #28, opened 2023-01-30
#24 Bump Kotlin and Compose

BorysChajdas:


Nataniel-Antosik (dismissed):

marcinWisnia (dismissed):

bartek977 (approved):

Nataniel-Antosik (approved):

app/build.gradle
```diff
@@ -48,7 +48,7 @@ android {
 
 dependencies {
 
-    def composeBom = platform('androidx.compose:compose-bom:2022.12.00')
+    def composeBom = platform('androidx.compose:compose-bom:2023.01.00')
     implementation composeBom
     androidTestImplementation composeBom
```
MarekMacko, line 53: Compose it's not used in instrumentation tests and it should be removed

build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
     id 'com.android.application' version '7.4.0' apply false
     id 'com.android.library' version '7.4.0' apply false
```
MarekMacko, line 3: Doesn't seem be required, we don't have any submodules, only the main "app"

app/build.gradle
BorysChajdas: Sure, I can handle it in next PR

build.gradle
BorysChajdas: I can handle it in next PR

PR #32, opened 2023-02-01
#20 Add ktlint

### LINKED ISSUE #20: Run ktlint on pr create

- [x] Add ktlint
- [x] Format code
- [x] Run ktlint on pr create

bartek977:


app/build.gradle.kts
```diff
@@ -64,4 +66,34 @@ dependencies {
...
+
+val ktlintFormat by tasks.creating(JavaExec::class) {
+    inputs.files(inputFiles)
+    outputs.dir(outputDir)
+
+    description = "Fix Kotlin code style deviations."
+    classpath = ktlint
+    mainClass.set("com.pinterest.ktlint.Main")
+    args = listOf("-F", "src/**/*.kt")
+}
```
bartek977: I've added ktlint configuration here as we have only one module at the moment, but when we add `domain` and `data` it will be better to extract this to a separate file, or I can do this now in a separate pr ;)

MarekMacko: Even if we have only one module, it should be placed in a separate file

app/build.gradle.kts
```diff
@@ -64,4 +66,34 @@ dependencies {
...
+
+    description = "Check Kotlin code style."
+    classpath = ktlint
+    mainClass.set("com.pinterest.ktlint.Main")
+    args = listOf("src/**/*.kt")
+}
+
+val ktlintFormat by tasks.creating(JavaExec::class) {
+    inputs.files(inputFiles)
+    outputs.dir(outputDir)
```
MarekMacko: Is it mandatory? Probably not so we can simplify it by getting rid of declaration of `outputDir` and `inputFiles`

MarekMacko (changes requested):

app/build.gradle.kts
bartek977: Extracted to separate file

app/build.gradle.kts
bartek977: Removed

settings.gradle.kts
```diff
@@ -14,3 +14,4 @@ dependencyResolutionManagement {
 }
 rootProject.name = "Picover"
 include(":app")
+include(":ktlint-ruleset")
```
MarekMacko: You can pass multiple projects to `include`

scripts/ktlint.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+val ktlintCheck by tasks.creating(JavaExec::class) {
+	description = "Check Kotlin code style."
+	classpath = ktlint
+	mainClass.set("com.pinterest.ktlint.Main")
+	args = listOf("src/**/*.kt")
+}
+
+val ktlintFormat by tasks.creating(JavaExec::class) {
```
MarekMacko: I'm always for a simplification, so do we really need `ktlintFormat`  in a new project?

bartek977: hmm, probably now the android studio formatter is enough, but in the future when we will add some custom ktlint rules then `ktlintFormat` will be helpful. Right now I remove this task.

settings.gradle.kts
bartek977: Done

MarekMacko (approved):

MarekMacko:
@michal-kucznerowicz we're waiting for you :)

michal-kucznerowicz (approved):

MarekMacko:
@bartek977 We're ready to go :)

PR #33, opened 2023-02-01
#20 Format code

### LINKED ISSUE #20: Run ktlint on pr create

bartek977:


app/src/main/java/com/example/android/picover/presentation/navigation/NavigationType.kt
```diff
@@ -3,6 +3,7 @@ package com.example.android.picover.presentation.navigation
 import androidx.compose.material3.windowsizeclass.WindowSizeClass
 import androidx.compose.material3.windowsizeclass.WindowWidthSizeClass
 
+@Suppress("ktlint:trailing-comma-on-declaration-site")
```
bartek977: What do you think about trailing comma? Here I had to add `@Suppress` annotation, because ktlint wanted to add the comma on the last enum item, then move the semicolon to the next line and finally throws error about unnecessary semi 😄

MarekMacko (approved):

michal-kucznerowicz (approved):

PR #34, opened 2023-02-01
#20 Run ktlint on pr create

### LINKED ISSUE #20: Run ktlint on pr create

bartek977:


michal-kucznerowicz (approved):

.github/workflows/default_branch_verification.yml
```diff
@@ -20,5 +20,7 @@ jobs:
 
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
+      - name: Analyze
+        run: ./gradlew ktlintCheck
```
Nataniel-Antosik: Could you look at the logs from action for example for this PR 
image
Does this indicate we have already `ktlint` in the `build`? If we will don't use the `build` for GitHub action and w use `assemble` then we should consider writing jobs for `ktlint`, `lint`, `tests`

Nataniel-Antosik: This logs was added when you create PR #32

.github/workflows/pull_request_verification.yml
```diff
@@ -20,5 +20,7 @@ jobs:
 
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
+      - name: Analyze
+        run: ./gradlew ktlintCheck
```
Nataniel-Antosik, line 24: Same situation as here: https://github.com/intive/Picover/pull/34/files#r1106782207

Nataniel-Antosik:
Minor: You named PR to add ktlint for creating PR and you add checking ktlint for Pull requests and also for branch develop.

bartek977:
> Minor: You named PR to add ktlint for creating PR and you add checking ktlint for Pull requests and also for branch develop.

fixed

.github/workflows/default_branch_verification.yml
bartek977: These logs are about building our `ktlint-ruleset` module where we will put custom rules

Nataniel-Antosik (approved):

PR #46, opened 2023-02-09
#24 Reformat code

BorysChajdas:
Intent changes due to our new .editorconfig

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,54 +1,54 @@
 plugins {
-    id("com.android.application")
-    id("org.jetbrains.kotlin.android")
+	id("com.android.application")
+	id("org.jetbrains.kotlin.android")
 }
 
 android {
-    namespace = "com.example.android.picover"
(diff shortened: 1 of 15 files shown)


michal-kucznerowicz:
In some cases, the indent is too large. It looks like 8, not 4 (which is recommended).

michal-kucznerowicz (changes requested):
In some cases, the indent is too large. It looks like 8, not 4 (which is recommended).

BorysChajdas:
@michal-kucznerowicz in which file for example?

michal-kucznerowicz:
In diff on github eg `PicoverNavigationBar` but after I checked it locally in AndroidStudio it looks good :D

michal-kucznerowicz (approved):

BorysChajdas:
@michal-kucznerowicz I think GitHub is displaying tabs and spaces differently that's why it looks like this 😄

Nataniel-Antosik (approved):

PR #47, opened 2023-02-09
#24 Add Hilt

BorysChajdas:


app/src/main/java/com/example/android/picover/presentation/navigation/PicoverNavHost.kt
```diff
@@ -24,10 +25,11 @@ fun PicoverNavHost(
 			HomeScreen()
 		}
 		composable("camera") {
-			CameraScreen()
+			CameraScreen(hiltViewModel())
```
michal-kucznerowicz: Minor: You could pass `hiltViewModel()` as the default argument in `CameraScreen`.

michal-kucznerowicz (approved):

app/build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 plugins {
```
marcinWisnia, line 1: @BorysChajdas  please remember that now we have "kotlin" name of packages, i see that PicoverApplication and CameraViewModel are in "java" named directory.

BorysChajdas: Yeah, rebase problem, thanks for catching it!

BorysChajdas: Done

marcinWisnia (approved):

PR #49, opened 2023-02-10
#41 Max line length in .editorconfig

### LINKED ISSUE #41: Max line length in .editorconfig
We should discuss which length will be appropriate in the `.editorconfig` file. So I leave comments ranging from 100 to the current value with step 10 points. 

**IMPORTANT** If you agree with concrete value leave a heart below comment

Nataniel-Antosik:
"Decided to go with 140 and 180 for tests."

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,7 +6,7 @@ insert_final_newline = true
 indent_style = tab
 indent_size = 4
 
-max_line_length = 180
+max_line_length = 140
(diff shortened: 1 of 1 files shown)


kmaslowiec (approved):
The idea to have 140 lines for coding and 180 seems legit.

MarekMacko (approved):

Issue #21, opened 2023-01-26
In settings or other place activity from Oss-licenses


BorysChajdas:
To be checked:
- a drawer in Compose on the last bottom nav item

If the drawer is difficult to achieve then the Oss will be placed on the More screen (Profile will be changed to More).

PR #51, opened 2023-02-14
#21 In settings or other place activity from Oss-licenses

marcinWisnia:
https://user-images.githubusercontent.com/49060907/218848693-c077eb62-dcbe-4f4b-8bb0-435107eaefc8.mp4

This is quite hacky :) modal navigation drawer is not a part of our navigation, it is shown when profile is clicked and then,  back button is overridden for closing it. It is also closed when we are changing the navigation item from the bar.

This solution applies only for navigation bar.

### DIFF
--- a/app/src/main/kotlin/com/example/android/picover/presentation/ProfileDrawerContent.kt
+++ b/app/src/main/kotlin/com/example/android/picover/presentation/ProfileDrawerContent.kt
@@ -1,0 +1,27 @@
+package com.example.android.picover.presentation
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.ModalDrawerSheet
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import com.intive.picover.R
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun ProfileDrawerContent() {
+	ModalDrawerSheet {
+		Box(
+			modifier = Modifier.fillMaxSize(),
+		) {
+			Text(
+				text = stringResource(R.string.Profile),
+				modifier = Modifier.align(Alignment.Center),
+			)
+		}
+	}
+}
--- a/app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavigationBar.kt
+++ b/app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavigationBar.kt
@@ -1,37 +1,60 @@
 package com.intive.picover.main.navigation.view
 
+import androidx.activity.compose.BackHandler
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.DrawerState
+import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
+import androidx.compose.material3.ModalNavigationDrawer
 import androidx.compose.material3.NavigationBar
 import androidx.compose.material3.NavigationBarItem
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.rememberCoroutineScope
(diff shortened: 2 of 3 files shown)


app/src/main/kotlin/com/example/android/picover/MainActivity.kt
```diff
@@ -32,18 +38,25 @@ class MainActivity : ComponentActivity() {
 					NavigationItem.Camera,
 					NavigationItem.Profile
 				)
-
 				Scaffold {
 					AnimatedVisibility(
 						visible = navigationType == NavigationType.NAVIGATION_BAR
 					) {
+						val coroutineScope = rememberCoroutineScope()
```
marcinWisnia: i am not a compose expert, and i am not sure how to handle with the coroutine scope across the nested elements, for example, if passing is it is a good practice ?

app/src/main/kotlin/com/example/android/picover/presentation/navigation/PicoverNavigationBar.kt
```diff
@@ -1,36 +1,61 @@
...
+				modifier = modifier.weight(1f),
+				navController = navController,
+				startDestination = NavigationItem.Home.route
+			)
+		}
 		NavigationBar(modifier = Modifier.fillMaxWidth()) {
 			items.forEach { item ->
 				NavigationBarItem(
-					selected = item.route == backStackEntry.value?.destination?.route,
+					selected = if (item is Profile) {
```
marcinWisnia: also here: extract such logic like this or it is better to have this like here :D

app/src/main/kotlin/com/example/android/picover/MainActivity.kt
michal-kucznerowicz: As far as I know, this is exactly the way how you should use it. Create a coroutine scope bound to the `AnimatedVisibility` composable lifecycle and create a new coroutine in its scope outside of a composable (e.g. in the `onClick()` lambda), so that it will be automatically canceled once it leaves the composition. To sum up `rememberCoroutineScope()` is designed to handle coroutine scope across nested elements. In another case, if you want to handle suspend functions inside a composable you should use `LaunchEffect`.

app/src/main/kotlin/com/example/android/picover/MainActivity.kt
```diff
@@ -32,18 +38,25 @@ class MainActivity : ComponentActivity() {
 					NavigationItem.Camera,
 					NavigationItem.Profile
```
michal-kucznerowicz: Remove the import of `Profile` or import all navigation items for consistency.

app/src/main/kotlin/com/example/android/picover/presentation/navigation/PicoverNavigationBar.kt
michal-kucznerowicz: For me, if it’s only UI related it can stay as is.

michal-kucznerowicz (changes requested):

app/src/main/kotlin/com/example/android/picover/MainActivity.kt
marcinWisnia: of course, fixed :) thanks !

marcinWisnia: @michal-kucznerowicz  please check my changes :)

michal-kucznerowicz (approved):

app/src/main/kotlin/com/example/android/picover/presentation/ProfileDrawerContent.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun ProfileDrawerContent() {
+	ModalDrawerSheet {
+		Box(
+			modifier = Modifier.fillMaxSize(),
+		) {
+			Text(
+				text = stringResource(id = R.string.Profile),
```
BorysChajdas: Minor: `id` named parameter could be dropped in that case as for `stringResource` it is obvious that we are passing string id 😄

BorysChajdas (approved):

app/src/main/kotlin/com/example/android/picover/presentation/ProfileDrawerContent.kt
marcinWisnia: naturlich :D changed :)

PR #57, opened 2023-02-19
#24 Basic camera usage

### LINKED ISSUE #24: Take a picture

BorysChajdas:


app/src/main/kotlin/com/intive/picover/camera/repository/ImageFileRepository.kt
```diff
@@ -0,0 +1,20 @@
...
+import java.io.File
+import javax.inject.Inject
+
+// TODO-BCH Should be unit tested after Kotest setup
+class ImageFileRepository @Inject constructor(
+	@ApplicationContext private val context: Context
+) {
+
+	fun createTempFileAndGetUri(): Uri? {
+		val tempFile = File.createTempFile("IMG_", ".jpg", context.getExternalFilesDir(Environment.DIRECTORY_PICTURES))
```
bartek977, line 17: it would be nice to extract these strings to vals or even create a file like Constants

app/src/main/kotlin/com/intive/picover/camera/view/CameraScreen.kt
```diff
@@ -1,24 +1,100 @@
...
+			alignment = Alignment.Center
+		),
+		previewPlaceholder = R.drawable.ic_launcher_background
+	)
+}
+
+@Composable
+private fun TakePictureButton(onClick: () -> Unit) {
+	Button(onClick = onClick) {
+		Text("Take a picture")
```
bartek977: shouldn't `Take a picture` come from resources? :D

app/src/main/res/xml/file_paths.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths>
+    <external-path
+        name="picoverImages"
+        path="Android/data/com.intive.picover/files/Pictures" />
```
bartek977: why not to use `ContextCompat.getExternalFilesDirs`? And I think we should consider if we want to share our photos with other apps (then probably we should save them in some public directory)

app/src/main/kotlin/com/intive/picover/camera/repository/ImageFileRepository.kt
BorysChajdas: What do we gain by doing so? It's used only in one place

bartek977: Okey, maybe I am too much sensitive on this :D But I bet `.jpg` will be used in the future not only once ;p For now as it's used only here it can be as it is

app/src/main/res/xml/file_paths.xml
BorysChajdas: I'm using `Context::getExternalFilesDirs` but the paths are needed. About sharing the photos with other apps – we will discuss it later, it's only a draft to start feature 😄

app/src/main/kotlin/com/intive/picover/camera/view/CameraScreen.kt
BorysChajdas: Right, changed, thanks!

app/src/main/res/xml/file_paths.xml
bartek977: I can't see where you're using `picoverImages`, could you explain what this path is needed for?

// Okey, I searched `picoverImages` but `file_paths` is used  :D

but do we need to specify the whole path? `Pictures/` wouldn't be enough?

BorysChajdas: Yeah, it is possible to simplify it a little bit using `external-files-path` instead of `external-path`. Have a look now!

bartek977 (approved):
please rebase

marcinWisnia (approved):

PR #62, opened 2023-03-09
#21 Add oss licenses plugin

marcinWisnia:


build.gradle.kts
```diff
@@ -1,3 +1,9 @@
+buildscript {
+	dependencies {
+		classpath("com.google.android.gms:oss-licenses-plugin:0.10.4")
```
marcinWisnia, line 3: Used 10.4 version to see the licences list :D (10.6 is not showing the list when the app is not in release version). If You think, that we should go with 10.6, i've created an issue for the update: [issue](https://github.com/orgs/intive/projects/3/views/1?pane=issue&itemId=22469292)

app/src/main/AndroidManifest.xml
```diff
@@ -28,6 +28,12 @@
                 android:name="android.app.lib_name"
                 android:value="" />
         </activity>
+        <activity
```
marcinWisnia, line 31: This oss activities needs Theme.appcompat kind of style to work. Now it is a little bit different from what we have for our compose elements in the app, but i've created an issue to handle this: [issue](https://github.com/orgs/intive/projects/3/views/1?pane=issue&itemId=22469670)

bartek977 (approved):

app/src/main/AndroidManifest.xml
Nataniel-Antosik: I agree with that. In the near future, we can think about changing the theme for this, or leaving it just like now because it's just a licenses list 😄

Nataniel-Antosik (approved):

app/src/main/AndroidManifest.xml
Nataniel-Antosik: I agree with that. In the near future, we can think about changing the theme for this, or leaving it just like now because it's just a licenses list 😄

app/src/main/kotlin/com/example/android/picover/presentation/ProfileDrawerContent.kt
```diff
@@ -1,14 +1,19 @@
 package com.example.android.picover.presentation
```
MarekMacko, line 1: @marcinWisnia Package should be corrected now

marcinWisnia: good catch :) thanks, will be fixed in upcoming PR :)

PR #71, opened 2023-03-24
#55 Find more secured way to store .json Firebase file

### LINKED ISSUE #55: Find more secured way to store .json Firebase file
I see now it's just thought and after our discussion about this I would like to take on this task :D ~Nataniel

Nataniel-Antosik:


.github/workflows/pull_request_verification.yml
```diff
@@ -21,7 +21,13 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Check lint
-        run: ./gradlew lintDebug
+        run: |
```
marcinWisnia: shouldn't this be extracted to another step ? is there a reason to keep this together with "Check lint" step ?

.github/workflows/pull_request_verification.yml
```diff
@@ -21,7 +21,13 @@ jobs:
...
+          gpg --quiet --batch --yes --decrypt --passphrase="$FIREBASE_SECRET" \
+          --output /home/runner/work/Picover/Picover/app/services.tar **/services.tar.gpg
+          tar xvf /home/runner/work/Picover/Picover/app/services.tar
```
marcinWisnia: shouldn't this by finally extracted to some file, which name is google-services.json and it is in the Picover/app/ directory ? As far as i know, it could be a problem to find it now during the build process

.github/workflows/pull_request_verification.yml
```diff
@@ -21,7 +21,13 @@ jobs:
...
+          ./gradlew lintDebug
+        env:
+          FIREBASE_SECRET: ${{ secrets.FIREBASE_SECRET }}
```
marcinWisnia: probably other way to handle this, but You probably noticed already :D : https://write.agrevolution.in/ci-android-part-1-964026a0356f

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: Yes, I can do this 😄

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I used way from the link which you sent in another comment

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: It works thanks 😎

MarekMacko:
What about running the app locally if `google-service.json` is missing?

Nataniel-Antosik:
Ok, @MarekMacko I will find a solution or solutions to have locally this file :D

marcinWisnia:
@Nataniel-Antosik any progress on this ? You need to clarify or discuss something here ?

Nataniel-Antosik:
@marcinWisnia I will work on it and I know It is `hanging` too long I hope I will add some ideas tomorrow to discuss or just decide what we will do 😄.

Nataniel-Antosik:
@intive/android-guild-szczecin 
My ideas:
- Storing this file in some place where everybody from our team can fetch it and use it during compile project.
- We can encrypt this file and hold it in the project therefore everybody can decrypt it and use it during compile project.
- Or we can just leave this file and do nothing with it (like some people on the internet advised) and just make the correct rules for our project on the firebase.

marcinWisnia:
@Nataniel-Antosik maybe You could try also solution with the firebase rules ? if this is also applicable for the access to the project ?

BorysChajdas (approved):

bartek977 (approved):

.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,10 @@ jobs:
 
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
+      - name: Decode google-service.json
```
BorysChajdas: Decode -> read

Nataniel-Antosik: Done

MarekMacko (approved):

MarekMacko:
@Nataniel-Antosik Ready to go!

PR #80, opened 2023-04-02
#77 Verify if the branch and commit satisfy template

### LINKED ISSUE #77: Verify if the branch and commit satisfy template
The job on GitHub fails if the branch or commit is in the wrong format (according to our readme).

Jobs should be independent (one for branch, one for commit).

Nataniel-Antosik:
## Checks
<p align="center"> 
	 
</p>

## Incorrect branch name
<p align="center"> 
	
</p>

## Incorrect commit message
<p align="center"> 
	
</p>

### DIFF
--- a/.github/workflows/pull_request_verification.yml
+++ b/.github/workflows/pull_request_verification.yml
@@ -5,10 +5,9 @@ on:
     branches: '**'
 
 jobs:
-  build:
 
+  setup_java:
     runs-on: ubuntu-latest
-
     steps:
       - uses: actions/checkout@v3
       - name: set up JDK 11
@@ -18,6 +17,34 @@ jobs:
           distribution: 'temurin'
           cache: gradle
 
+  branch_name:
+    needs: setup_java
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Give Permission to File checkBranch.sh
+        run: chmod +x ./.github/scripts/branch/checkBranch.sh
+      - name: Check correct branch name
+        run: ./.github/scripts/branch/checkBranch.sh
+
+  commit_message:
+    needs: setup_java
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{github.event.after}}
+      - name: Give Permission to File checkCommitMessage.sh
+        run: chmod +x ./.github/scripts/commit/checkCommitMessage.sh
+      - name: Check correct commit message
+        run: ./.github/scripts/commit/checkCommitMessage.sh
+
+  build:
+    needs: setup_java
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Check lint
--- a/.github/scripts/branch/checkBranch.sh
+++ b/.github/scripts/branch/checkBranch.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+branch=${GITHUB_HEAD_REF#refs/*/v}
+if ! [[ $branch =~ [a-z]{3}\/[0-9]+-[A-Za-z-]+$ ]]
+then
+    echo "::error::Branch is invalid — $branch"
+    echo "::error::Branch should looks like - \$acronym/\$issue_number-\$description"
+    exit 1
+fi
--- a/.github/scripts/commit/checkCommitMessage.sh
+++ b/.github/scripts/commit/checkCommitMessage.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+message=$(git log -1 --pretty=format:"%s" | tr -dc '[:print:]\n')
+if ! [[ "$message" =~ ^#[0-9]+[[:space:]][^[:cntrl:]]+[[:space:]]*$ ]]
+then
+  echo "::error::Message is invalid — $message"
+  echo "::error::Commit message should look like - #\$issue_number \$description"
+  exit 1
+fi


Nataniel-Antosik:
## Incorrect branch
<p align="center"> 
	 
</p>

## Incorrect commit message
<p align="center"> 
	
</p>

Nataniel-Antosik:
I prepared a branch to show how errors will be looks like for incorrect branch name and commit message. #81

.github/workflows/pull_request_verification.yml
```diff
@@ -5,10 +5,59 @@ on:
...
+      - name: Check correct branch name
+        run: |
+          tag=${GITHUB_HEAD_REF#refs/*/v}
+          if [[ $tag =~ [a-z]{2,}\/[0-9]+-[A-Za-z-]+$ ]]
+          then
+            echo "::notice::Tag is valid — $tag"
+            echo "{tag}={GITHUB_HEAD_REF#refs/*/v}" >> $GITHUB_OUTPUT
+          else
+            echo "::error::Tag is invalid — $tag"
+            echo "::error::Tag should looks like - \$acronym/\$issue_number-\$description"
```
bartek977: shouldn't be branch instead of tag?

.github/workflows/pull_request_verification.yml
```diff
@@ -5,10 +5,59 @@ on:
...
+            exit 1
+          fi
```
bartek977: is it possible to easily extract this check expression to some function/script?

Nataniel-Antosik: I will check it

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I think it could be a `branch`

.github/workflows/pull_request_verification.yml
```diff
@@ -5,10 +5,59 @@ on:
...
+          distribution: 'temurin'
+          cache: gradle
+
+      - name: Check correct branch name
+        run: |
+          tag=${GITHUB_HEAD_REF#refs/*/v}
+          if [[ $tag =~ [a-z]{2,}\/[0-9]+-[A-Za-z-]+$ ]]
+          then
+            echo "::notice::Tag is valid — $tag"
+            echo "{tag}={GITHUB_HEAD_REF#refs/*/v}" >> $GITHUB_OUTPUT
```
marcinWisnia: i am wondering if we really need message that the branch name is valid, maybe we should only receive message when the branch name is invalid.

.github/workflows/pull_request_verification.yml
```diff
@@ -5,10 +5,59 @@ on:
...
+          java-version: '11'
+          distribution: 'temurin'
+          cache: gradle
 
+      - name: Check correct commit message
+        run: |
+          message=$(git log -1 --pretty=format:"%s" | tr -dc '[:print:]\n')
+          if [[ "$message" =~ ^#[0-9]+[[:space:]][^[:cntrl:]]+[[:space:]]*$ ]]
+          then
+            echo "::notice::Message is valid — $message"
```
marcinWisnia: Same as for branch

.github/workflows/pull_request_verification.yml
```diff
@@ -5,10 +5,59 @@ on:
...
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: ${{github.event.after}}
+      - name: set up JDK 11
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          cache: gradle
```
marcinWisnia: i am also wondering, if there is a way to extract common part of the jobs, for example "set up jdk 11"

Nataniel-Antosik: I extracted this to the independent job and others are dependent on this.

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I deleted this message.

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I deleted this message.

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I created a file for this script.

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: I changed it.

.github/workflows/checkBranch.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+branch=${GITHUB_HEAD_REF#refs/*/v}
+if ! [[ $branch =~ [a-z]{2,}\/[0-9]+-[A-Za-z-]+$ ]]
```
bartek977: acronym can have unlimited characters? I remember something about max 3 characters, but it's possible that this requirement comes from another project 😄

Nataniel-Antosik: Yes, it should be exactly 3 characters. I changed it. 😄

bartek977 (approved):

.github/workflows/checkBranch.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
MarekMacko: Should it be placed under `/workflows/` directory? If we would add more scripts here, then we will end up with a mess. Try to group them, use the same rule as for source files `.kt` or `.java`

Nataniel-Antosik: Yes, you have a point about the potential mess in directory `workflows`. These two scripts are related to GitHub action so I will create a directory `bash` in `.github` also in the new directory I will group these two scripts to correct directory groups like `branch`, and `commit`. What do you think about this?

MarekMacko: Just add your proposal and we will see ;)

Nataniel-Antosik: Ok, I added 😄

.github/bash/branch/checkBranch.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
MarekMacko: If we would change the file format bash script to python, then would we need to rename the directory from `bash` to `python`?  :)

Nataniel-Antosik: Yes, it could be a little problem in the future. So I changed the directory name from `bash` to `scripts`. 😄

marcinWisnia (approved):

MarekMacko (approved):

Issue #65, opened 2023-03-09
Add authentication


marcinWisnia:
Authentication will be done via firebase -> to precise, what kind of auth we want (google/mail).

Authentication is required to use the app. In the case of a not logged user, the authentication screen should be displayed.

The logout option should be added to the profile.

PR #84, opened 2023-05-10
#65 Update Gradle

BorysChajdas:


app/build.gradle.kts
```diff
@@ -42,6 +42,10 @@ android {
 			excludes += "/META-INF/{AL2.0,LGPL2.1}"
 		}
 	}
+	compileOptions {
+		sourceCompatibility = JavaVersion.VERSION_17
+		targetCompatibility = JavaVersion.VERSION_17
+	}
```
MarekMacko, line 48: Try to apply the same settings for lint module

BorysChajdas: Actually, it is not a problem with a Gradle configuration – locally everything is working fine. The problem is with our Github Actions configuration, the setup java is wrong configured so the build action is using the default Java from the Linux distro (which is 11, that's why it was working before 😄).

BorysChajdas: @MarekMacko Now it works! The problem lay elsewhere than we expected 😄

MarekMacko: Always expect unexpected 😄

MarekMacko (approved):

marcinWisnia (approved):

PR #88, opened 2023-05-11
#65 Update CI Java

BorysChajdas:


.github/workflows/default_branch_verification.yml
```diff
@@ -12,10 +12,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: set up JDK 11
```
MarekMacko, line 15: Since it's only a name and to avoid unecessary changes, can we drop the value and just keep simple "set up JDK"?

MarekMacko (approved):

.github/workflows/default_branch_verification.yml
BorysChajdas: Right, done!

MarekMacko (approved):

Nataniel-Antosik (approved):

Issue #64, opened 2023-03-09
Optimise app builds on CI


marcinWisnia:
## Description
- repeating `chmod +x gradlew` in each job is redundant, try to find a way how to get rid of it,
- run locally `./gradlew build --dry-run` to check which tasks are executed during `build` command, 
  - do we need all of them?
  - should we keep all of them in one job?

### Make more readable regex
Written `regex` after time could be vague. So we should make it more readable.

### Optimize CI
At this moment our `build job` executed more than ten minutes it should be faster.

### Add checking count of commits 
We have the rule one `pull request` one `commit` therefore we should have a job that will check if a pull request has one commit.

### Helpful Links: 
- [DRY: REUSING CODE IN GITHUB ACTIONS](https://cardinalby.github.io/blog/post/github-actions/dry-reusing-code-in-github-actions/)

## Tasks
- [x] Make more readable regex
- [x] Optimize CI
- [x] Add checking count of commits 
- [x] Use Docker image
- [x] Fix checking commit message

BorysChajdas:
chmod +x on each job removed in: https://github.com/intive/Picover/pull/90

PR #90, opened 2023-05-12
#65 Fix actions to use correct java

BorysChajdas:
Connected with #64, as it removed granting the permission in each run.


BorysChajdas:
Setup Java was wrongly used in our configuration. setup_java job result was not used, `needs: job` only checks that the job will wait for the other job to finish, but our other jobs were not using the java configured by setup_java.

.github/workflows/pull_request_verification.yml
```diff
@@ -1,52 +1,38 @@
...
       - name: Check correct commit message
         run: ./.github/scripts/commit/checkCommitMessage.sh
 
   build:
-    needs: setup_java
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - name: Grant execute permission for gradlew
-        run: chmod +x gradlew
```
MarekMacko, line 49: Finally done! 😄 Please add a reference to #64 - it fixes one topic

MarekMacko (approved):

.github/workflows/pull_request_verification.yml
```diff
@@ -1,52 +1,38 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
```
MarekMacko, line 34: Is there a way to extract it, not repeat the same setup for almost each job? Something like tamplate for Gitlab job

BorysChajdas: Yeah, I asked myself the same question as I looked at it 😄 at the moment I do not know, because I wanted to fix what is blocking us, but I will still look at it

Nataniel-Antosik (approved):

PR #93, opened 2023-05-18
#65 Use existing profile screen instead of ProfileDrawerContent

BorysChajdas:


app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -1,22 +1,32 @@
...
 		modifier = Modifier.fillMaxSize(),
+		horizontalAlignment = Alignment.CenterHorizontally,
 	) {
-		Text(
-			text = stringResource(id = R.string.Profile),
-			modifier = Modifier.align(Alignment.Center),
-		)
+		Button(
+			onClick = {
+				ContextCompat.startActivity(context, Intent(context, OssLicensesMenuActivity::class.java), null)
```
MarekMacko: What is the point to use `ContextCompat.startActivity` instead of regular `context.startActivity`?

BorysChajdas: Probably by mistake, I've copied it from the current implementation. Following your question I changed to regular `startActivity` and it works like before.

MarekMacko: 😞

MarekMacko (approved):

marcinWisnia (approved):

PR #96, opened 2023-05-19
#65 hiltLifecycleViewModel to automatically observe lifecycle

BorysChajdas:


app/src/main/kotlin/com/intive/picover/common/lifecycle/HiltLIfecycleViewModel.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.hilt.navigation.compose.hiltViewModel
+import com.intive.picover.common.viewmodel.LifecycleViewModel
+
+@Composable
+inline fun <reified VM : LifecycleViewModel> hiltLifecycleViewModel() =
+	hiltViewModel<VM>().also {
+		LocalLifecycleOwner.current.lifecycle.addObserver(it)
+	}
```
MarekMacko, line 12: Would be nice to see some tests :)

BorysChajdas: Unit test `Composable`? 😮

MarekMacko: Why not? 😮

BorysChajdas: Because to test it the test functions would need to be `Composable`

BorysChajdas: also the `hiltViewModel` is an inline function so I don't think that there is any way to mockk it

MarekMacko: Sure, but there is still a way to do so ;)

BorysChajdas:
Closing because we decided to not use `LifecycleViewModel` at all

PR #102, opened 2023-05-23
#65 Auth repository implementation

BorysChajdas:


app/src/main/kotlin/com/intive/picover/auth/di/AuthModule.kt
```diff
@@ -0,0 +1,17 @@
...
+import com.google.firebase.auth.ktx.auth
+import com.google.firebase.ktx.Firebase
+import dagger.Module
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.components.SingletonComponent
+
+@Module
+@InstallIn(SingletonComponent::class)
+object AuthModule {
```
bartek977, line 12: will be here some additional providers later? if not maybe you could rename it to `AppModule` as you suggested in another pr?

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+					AuthEvent.Logged
+				}.let { authEvent ->
+					trySend(authEvent)
+				}
+			}
+			firebaseAuth.addAuthStateListener(authStateListener)
+			awaitClose { firebaseAuth.removeAuthStateListener(authStateListener) }
+		}
+
+	fun logout() {
```
bartek977, line 29: shouldn't be suspend fun and moved to dispatcher io?

BorysChajdas: Maybe, I don't know 😄, unlike `firebaseAuth.currentUser.delete()` the `firebaseAuth.signOut()` is not even returning `Task` so I suppose that there is no need to make it suspend. But maybe you see any reason here?

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+		firebaseAuth.signOut()
+	}
+
+	// TODO #100 Handle FirebaseAuthRecentLoginRequiredException
+	suspend fun deleteAccount() {
+		firebaseAuth.currentUser!!
```
kmaslowiec, line 35: Is !! the operator necessary? Do we have an option to avoid it?

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.coroutines.tasks.await
+
+class AuthRepository @Inject constructor(
+	private val firebaseAuth: FirebaseAuth,
+) {
+
+	fun observeEvents() =
+		callbackFlow {
+			val authStateListener = FirebaseAuth.AuthStateListener {
+				if (it.currentUser == null) {
```
Nataniel-Antosik, line 17: minor: `it` could be with a different title

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
BorysChajdas: Of course, we can avoid it and use the safe call `?.` but it could lead to a hidden not working option (nothing will happen after clicking, but the app will not crash). Because of that I decided to force non-nullable here, and the app should crash when the user is trying to delete the account but Firebase does not have currentUser. Thanks to that we can track all edge-cases.

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
BorysChajdas: Will it improve readability in your opinion? Because for me it looks redundant to use named receiver is such cases :D

bartek977 (approved):

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
Nataniel-Antosik: Sometimes for me is more readable but I know it depends on the situation when it should be named differently than `it`. Below you named `authEvent` probably instead of `it2` I think 😅. So it is not very important to me so if you would like to leave it like it is now, I am ok with it 😄.

Nataniel-Antosik (approved):

PR #104, opened 2023-05-23
#65 Implement authentication feature

BorysChajdas:


app/src/test/kotlin/com/intive/picover/main/viewmodel/MainViewModelTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+class MainViewModelTest : ShouldSpec(
+	{
+		val authRepository: AuthRepository = mockk()
+		val tested by lazy { MainViewModel(authRepository) }
+
+		should("set state according to AuthEvent returned from observeEvents WHEN created") {
+			every { authRepository.observeEvents() } returns flowOf(AuthEvent.Logged)
+
+			tested.state.first() shouldBe MainState.UserAuthorized
```
bartek977, line 21: what about `AuthEvent.NotLogged`? ;p

bartek977: shouldn't be it compared to value from `MainState.fromAuthEvent()`?

BorysChajdas: I decided to simplify it because `MainState::fromAuthEvent` is separately tested. But probably you are right, I will adjust the test. Thanks!

BorysChajdas: Done, I've mocked the response from `MainState::fromAuthEvent` as it is already tested.

kmaslowiec (approved):

bartek977 (approved):

Issue #100, opened 2023-05-23
Handle FirebaseAuthRecentLoginRequiredException when deleting an account


BorysChajdas:
When the user authenticated into the app some time ago the delete account functionality did not work and is crashing the app because Firebase is throwing `FirebaseAuthRecentLoginRequiredException`:
```
com.google.firebase.auth.FirebaseAuthRecentLoginRequiredException: This operation is sensitive and requires recent authentication. Log in again before retrying this request.
```
To support account deletion using Firebase Auth we need to redirect the user to log in again before deleting an account.

MarekMacko:
@BorysChajdas  Please close it if there is nothing to do anymore

PR #105, opened 2023-05-23
#65 Add logout and delete account on profile

BorysChajdas:


kmaslowiec (approved):

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -0,0 +1,24 @@
...
+) : ViewModel() {
+
+	fun onLogoutClick() {
+		authRepository.logout()
+	}
+
+	fun onDeleteAccountClick() {
+		viewModelScope.launch {
+			authRepository.deleteAccount()
+		}
```
bartek977, line 22: what happened if some error occurs? for example there will be no internet connection?

BorysChajdas: The app will crash 😄 for now I've not added any error handling as you can see. I decided to skip it for now, and it will be a part of another task.

BorysChajdas: Currently, that is only working for a happy-path but I think that's enough for now. I've created one issue #100 which also needs to be handled because account deletion is only possible with "fresh login credentials" so if you are logged for example for 2 days and try to delete your account the app will crash because according to Firebase we should first redirect the user to log-in again to make sure that the credentials are fresh and the user is 100% sure that he wanted to delete an account.

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -0,0 +1,24 @@
...
+import javax.inject.Inject
+import kotlinx.coroutines.launch
+
+@HiltViewModel
+class ProfileViewModel @Inject constructor(
+	private val authRepository: AuthRepository,
+) : ViewModel() {
+
+	fun onLogoutClick() {
+		authRepository.logout()
```
kmaslowiec, line 16: It seems that the logout does not return us to the sign-in page, only deleting the account does:

https://github.com/intive/Picover/assets/33199010/f25091dc-7105-4447-b422-c50150430d9c

bartek977: @BorysChajdas will it be done in this pr or another one?

BorysChajdas: That's not an issue of the implementation but the fact that you saved the credentials using "Smart Lock". Without saving the password it works as expected, however when the password is saved to "Smart Lock" then after logout you are automatically logged in. We can think of it later, maybe there is some way to tweak it, or we can disable "Smart Lock" (I've seen option for that).

kmaslowiec: Ok. Good to know. Thanks.

marcinWisnia (approved):

PR #111, opened 2023-05-31
#65 Bump AGP to 8.0.2

BorysChajdas:


BorysChajdas:
@Nataniel-Antosik I think we should adjust the branch-name script to allow numbers in the branch name, as you can see it is failing for "bch/65-bump-agp-to-8.02" but the commit-message is passing.

Nataniel-Antosik:
@BorysChajdas I will fix it ASAP

PR #114, opened 2023-05-31
#64 Adjust branch name regex 1.0.1

### LINKED ISSUE #64: Optimise app builds on CI

Nataniel-Antosik:


BorysChajdas (approved):

.github/scripts/branch/checkBranch.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 branch=${GITHUB_HEAD_REF#refs/*/v}
-if ! [[ $branch =~ [a-z]{3}\/[0-9]+-[A-Za-z-]+$ ]]
+if ! [[ $branch =~ [a-z]{3}\/[0-9]+-[^[:cntrl:]]+$ ]]
```
BorysChajdas, line 4: To rethink: Probably we should not allow ALL characters here, for example new line or some strange characters should not be allowed

Nataniel-Antosik: We can think about it in the near future. Thanks to creating a new branch from AS we can't add blank space because AS adds character `_` therefore some cases is not possible 😄

marcinWisnia (approved):

Issue #79, opened 2023-03-30
Fetch list of parties from API


BorysChajdas:
Use Firebase

PR #117, opened 2023-06-07
#79 Add kotlin reflect

marcinWisnia:
One and only solution I've found for having a possibility to trigger single test when NOT using plugin "window"

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -77,5 +77,8 @@ dependencies {
 	debugImplementation(libs.compose.ui.tooling)
 	testImplementation(libs.test.kotest.runner)
 	testImplementation(libs.test.mockk)
+	testCompileOnly("org.jetbrains.kotlin:kotlin-reflect:1.8.20") {
+		because("Needed to locally trigger single kotest test - check new versions of kotlin and kotest plugins to fix this workaround")
+	}
 }
 


BorysChajdas (approved):

bartek977 (approved):

MarekMacko:
What is the reason of that change?

marcinWisnia:
> What is the reason of that change?

Because when executing single test using play icon, we are getting: 
java.lang.NoClassDefFoundError: kotlin/reflect/full/KClasses and this is the only solution i’ve found for getting this working

marcinWisnia:
> What is the reason of that change?

Because when executing single test using play icon, we are getting: 
java.lang.NoClassDefFoundError: kotlin/reflect/full/KClasses and this is the only solution i’ve found for getting this working

app/build.gradle.kts
```diff
@@ -82,5 +82,6 @@ dependencies {
 	testImplementation(platform("io.kotest:kotest-bom:5.6.2"))
 	testImplementation("io.kotest:kotest-runner-junit5")
 	testImplementation("io.mockk:mockk:1.13.5")
+	testImplementation("org.jetbrains.kotlin:kotlin-reflect:1.8.21")
```
MarekMacko: According to https://github.com/intive/Picover/pull/117#issuecomment-1592810548, if dependency is only added to solve the issue of running the tests, which is strictly connected with the kotest plugin, then we shouldn't have an access to classes in test source files. Try to use `testRuntime` or `testCompile` for that. We shouldn't expose too much. <br>
Would be also nice to put comment of whatsoever, it's rather a temporary workaround than a regular dependency needed for test execution, e.g. on CI it's not needed. Dependency notation can have additional "because" field with explanation, it would fit our needs here :)

marcinWisnia: Oh yes, You are right in this case :D

I've checked testRuntime and it is not working in our case, but what is working, is `testCompileOnly`, but i am receiving:

`> Cannot find a version of 'org.jetbrains.kotlin:kotlin-reflect' that satisfies the version constraints:
           Dependency path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:1.8.21' because of the following reason: Needed to locally trigger single kotest test
           Constraint path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:{strictly 1.8.20}' because of the following reason: debugUnitTestRuntimeClasspath uses version 1.8.20`

it looks few other dependencies is also using reflect which is problematic in this case...I've tried:
- using `strictly` with 1.8.21 specified version -> not working
- excluding kotlint reflect from dependencies using it in their code -> works but we have to do it for every library
- use `resolutionStrategy` in `configurations.all` with forcing this dependency and its specific version -> works

So i think that options are going with what i prepared with adding reason as a `because` or use `testCompileOnly`with `resolutionStrategy`

MarekMacko: `using strictly with 1.8.21 specified version -> not working` but the error says about using version 1.8.20. 
What does it mean "not working" in this case?

marcinWisnia: @MarekMacko I mean that I used:


`testCompileOnly("org.jetbrains.kotlin:kotlin-reflect"){
         version {
             strictly("1.8.21")
         } 
}`

when declaring dependency and received exactly the same error mentioned in previous comment. Looks like other dependencies uses reflect, with lower version (1.8.20 or lower -> and in some of them the version is declared using `strictly`, because i saw this when i was printing dependencies tree). 

So I think we need forcing this version of dependency, with resolutionStrategy, to go with usage of `testCompileOnly`.
Otherwise, i will go with the code i currently have on this PR (with `testImplementation` usage but with `because` description)

MarekMacko: Error says to use version `1.8.20`, not `1.8.21`.
I've checked with version `1.8.20` and it works on my side
```
	testCompileOnly(libs.test.kotlin.reflect) {
		version { strictly("1.8.20") }
	}

marcinWisnia: yes, i know that this solution would work :D 

But the question is if we want to use older version, instead of our actual kotlin version :D I think that we should use one version, for reflect and kotlin across all project rather than forcing here "lower" dependency version :D 

That's the reason why i was so stubborn :D

MarekMacko: But where we are using it? We don't have an access to anything from `kotlin-reflect`. 
Is there any reason to expose more than it's needed? This declaration is only a temporary workaround, not a solution to stay or am I wrong? 
`testImplementation` exposes dependency to runtime, so the test execution can be affected. Such workarounds are hard to maintain in long term and nobody will remember about.

marcinWisnia: ok, as it is a workaround, adjusted now :) thanks for Your assistance !

MarekMacko (approved):

PR #118, opened 2023-06-07
#79 Fetch parties from firebase

marcinWisnia:


app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
```diff
@@ -4,4 +4,6 @@ data class Party(
 	val id: Int,
 	val title: String,
 	val description: String,
-)
+) {
+	constructor() : this(0, "", "")
```
marcinWisnia: there is an error, when using object without default values, so used this approach. Alternative is use: val id: Int = 0 etc

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,29 @@
...
-		Party(id = 5, title = "title5", description = "description5"),
-	)
+	private val databaseReference =
+		firebaseDatabase.getReference("Parties")
 
-	fun getParties() = parties
+	fun parties() =
+		databaseReference.snapshots.map { snapshot ->
+			snapshot.children.map {
+				it.getValue(Party::class.java)!!
```
marcinWisnia: there is also an ktx extension for this, like getValue<>Party()  but couldn't struggle through the tests :/

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,29 @@
...
+			}
+		}
 
-	fun getPartyById(id: Int) = parties.find { it.id == id }!!
+	fun partyById(id: Int) =
```
marcinWisnia, line 23: of course this is TEMPORARY -> this will be handled in different way, when we will have adding to the firebase, and then, we could observe and access only one desired object (which will be created with creating key and key will be also used as an ID) so the path to one object will be accessed and observed

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -3,23 +3,38 @@ package com.intive.picover.parties.viewmodel
...
 	init {
-		loadParties()
+		viewModelScope.launch {
+			loadParties()
+		}
 	}
 
-	private fun loadParties() {
-		parties.value = PartiesState.Loaded(partiesRepository.getParties())
+	private suspend fun loadParties() {
```
marcinWisnia: good question how we would like to handle this in viewmodels, we can do it like this, OR:   

Use flow here, and for example in tests, test happy path, with first and second flow value
and test error path with first and second value of flow (using Turbine for example: https://github.com/cashapp/turbine)

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,29 @@
...
-		Party(id = 3, title = "title3", description = "description3"),
-		Party(id = 4, title = "title4", description = "description4"),
-		Party(id = 5, title = "title5", description = "description5"),
-	)
+	private val databaseReference =
+		firebaseDatabase.getReference("Parties")
 
-	fun getParties() = parties
+	fun parties() =
+		databaseReference.snapshots.map { snapshot ->
```
marcinWisnia: of course alternative is firestore, read guide and decided to use this one (realtime database) :D

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,29 @@
...
+import com.google.firebase.database.FirebaseDatabase
+import com.google.firebase.database.ktx.snapshots
 import com.intive.picover.parties.model.Party
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.flow.map
 
 @Singleton
-class MockedPartiesRepository @Inject constructor() {
+class MockedPartiesRepository @Inject constructor(
```
marcinWisnia, line 9: Rename will be added in next pr to not loose changes history

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
MarekMacko: What is the origin of the error?

app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
```diff
@@ -10,9 +12,13 @@ import kotlinx.coroutines.Dispatchers
 @Module
 @InstallIn(SingletonComponent::class)
 object AppModule {
+
 	@Provides
 	fun provideFirebaseStoreInstance() = FirebaseStorage.getInstance()
 
+	@Provides
+	fun provideFirebaseDatabase() = Firebase.database
```
MarekMacko, line 20: I think we should watch out with placing everything in `AppModule`. It creates bad habit to place everything in one module, in Singleton component. In the end, such habit with be followed by the rest of the team, just because "it doesn't work with different scope/component". Lifecycle of that instance should be relaxed.

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,29 @@
...
+) {
 
-	private val parties = listOf(
-		Party(id = 1, title = "title1", description = "description1"),
-		Party(id = 2, title = "title2", description = "description2"),
-		Party(id = 3, title = "title3", description = "description3"),
-		Party(id = 4, title = "title4", description = "description4"),
-		Party(id = 5, title = "title5", description = "description5"),
-	)
+	private val databaseReference =
```
MarekMacko: It's reather "parties" than general "database"

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
MarekMacko: What was the error?

app/src/main/kotlin/com/intive/picover/parties/state/PartiesState.kt
```diff
@@ -5,4 +5,5 @@ import com.intive.picover.parties.model.Party
 sealed class PartiesState {
 	object Loading : PartiesState()
 	data class Loaded(val parties: List<Party>) : PartiesState()
+	object Error : PartiesState()
```
MarekMacko: I think it's time create some base state, just to follow basie DRY rule, and not create same state types for each screen

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
marcinWisnia: The error is:
`com.google.firebase.database.DatabaseException: Class com.intive.picover.parties.model.Party does not define a no-argument constructor.` It looks like firebase need a non-arg constructor

I see another interesting solution for that -> use custom annotation class together with usage of No-arg compiler plugin ( https://kotlinlang.org/docs/no-arg-plugin.html)

I’ve done some quick tests and looks like it works (just annotate the data class with custom anotation and no need to use no-argument constructor or default params)

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
marcinWisnia: i am getting:

`Caused by: io.mockk.MockKException: no answer found for DataSnapshot(#5).getValue(com.intive.picover.parties.repository.MockedPartiesRepository$parties$lambda$1$lambda$0$$inlined$getValue$1@652f56c) among the configured answers: (
DataSnapshot(#5).getValue(com.intive.picover.parties.repository.MockedPartiesRepositoryTest$1$3$partySnapshot$1$1$invoke$$inlined$getValue$1@128b48f3))))`

Maybe it is because the getValue ktx extension is inlined: 

`inline fun <reified T> DataSnapshot.getValue(): T? {
    return getValue(object : GenericTypeIndicator<T>() {})
}`

Was thinkig also about wrapping this function, but i am not sure if it makes sense…

app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
marcinWisnia: Yes, i agree.

But i am wondering, if we should create a separate, firebasse  module, or firebaseDatabase and firebaseStorage modules, as we have separate module for auth. 

I am not sure about the correct path of growing on this level of implementation

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
marcinWisnia: Yeah right, leftover -> fixed !

app/src/main/kotlin/com/intive/picover/parties/state/PartiesState.kt
marcinWisnia: fully agree :D created issue for that, and started working on it: #127  (will add in separate pr)

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
MarekMacko: The same model for UI and Firebase response!?

app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
BorysChajdas: I think that even without our `SingletonComponent` module the Firebase is a singleton, and every call to the `Firebase.database` will provide us with the same instance. Fully agree with the fact that we should not put everything in the app module, but in my opinion, in the case of such dependencies like Firebase, it's OK to put them in the app module. Is there any benefit of extracting it to a separate module, or making it for example `ViewModelComponent`? Another thing is that it will be easier in the future to split the `AppModule` into a few other modules than to combine 5 different redundant modules with 1-2 provides into one. If we are not sure how we want to divide such dependencies, it is better to start with the simplest solution.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -27,18 +29,22 @@ fun PartiesScreen(
...
 			LoadedContent(
 				parties = state.parties,
 				openDetails = {
 					navController.navigate("partyDetails/$it")
 				},
 			)
 		}
+
+		is Error -> {
+			// To be implemented
```
BorysChajdas: Would be nice to add `TODO` to easily find such things later.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -3,23 +3,38 @@ package com.intive.picover.parties.viewmodel
...
+		runCatching {
```
BorysChajdas: I think that `runCatching` is wrongly used here, because we are only checking exception of the `partiesRepository.parties()` but we should check exceptions which could be thrown during Flow collection, so it should be rather something like that:
```
		runCatching {
			partiesRepository.parties().collect {
				parties.value = Loaded(it)
			}
		}.onFailure {
			parties.value = Error
		}
```

app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
MarekMacko: I wouldn't agree, nice example of messing `AppModule`, is our project. For everyone it's easier to put everything in one `SingletonComponent` module, because it's the easiest way to make it work without any errors, but also without thinking what component and scoping should be used.  
Indeed, it will be easier to combine 5 different modules, with different components and scoping, than splitting big `AppModule`, and most likely nobody would like to handle such splitting because it's a painful when such module will grow too much. 
I would not block current approach, it's not a big mistake, but for me it's not appropriate in a long term.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: ...make sense :D i am wondering about the types of exception that could be thrown by the collection itself and when they could be thrown, but i think this is something to dive deeper in :D

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
marcinWisnia: Not at all :D i left this intentionally to stimulate some of team members to discuss this, as our review is quite dead :D I've added proposal that i mentioned on one of guild meetings -> remote model and UI model

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
marcinWisnia: yes, forgot about -> extracted !

Nataniel-Antosik:
Please rebase the branch to take a new CI 😄

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
BorysChajdas: I think that all of the exceptions will be thrown during the collection for example when there is no internet connection, without the `collect` there will be no exception, because the call will not start :D

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
```diff
@@ -1,21 +1,27 @@
...
-		Party(id = 5, title = "title5", description = "description5"),
-	)
+	private val partiesReference =
+		firebaseDatabase.getReference("Parties")
 
-	fun getParties() = parties
+	fun parties() =
+		partiesReference.snapshots.map { snapshot ->
+			snapshot.children.map {
+				it.getValue(PartyRemote::class.java)!!
```
Nataniel-Antosik: You used `snapshot` above and here `it`, IMO to save consistency you can name it correctly to avoid using `it` in this case

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -27,18 +29,22 @@ fun PartiesScreen(
...
 			LoadedContent(
 				parties = state.parties,
 				openDetails = {
 					navController.navigate("partyDetails/$it")
 				},
 			)
 		}
+
+		is Error -> {
+			// TODO: Will be implemented in #133
```
Nataniel-Antosik, line 40: minor: You added here comment between `{ }` instead of doing the same as in `PartyDetailsScreen.kt` or vice versa

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -3,23 +3,37 @@ package com.intive.picover.parties.viewmodel
...
 class PartiesViewModel @Inject constructor(
 	private val partiesRepository: MockedPartiesRepository,
 ) : ViewModel() {
 
-	val parties: MutableState<PartiesState> = mutableStateOf(PartiesState.Loading)
+	val parties: MutableState<PartiesState> = mutableStateOf(Loading)
 
 	init {
-		loadParties()
+		viewModelScope.launch {
```
Nataniel-Antosik: I am not sure `launch` from `viewModelScope` should be here maybe we can use it in `loadParites()`

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartyDetailsViewModel.kt
```diff
@@ -4,25 +4,39 @@ import androidx.compose.runtime.MutableState
...
 ) : ViewModel() {
 
-	val partyDetails: MutableState<PartyDetailsState> = mutableStateOf(PartyDetailsState.Loading)
+	val partyDetails: MutableState<PartyDetailsState> = mutableStateOf(Loading)
 	private val partyId: Int = savedStateHandle["partyId"]!!
 
 	init {
-		loadParty(partyId)
+		viewModelScope.launch {
+			loadParty(partyId)
```
Nataniel-Antosik: same like here: [PartiesViewModel.kt](https://github.com/intive/Picover/pull/118#discussion_r1242079501)

app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
BorysChajdas: So I think it's a nice candidate to encourage all team members to think about that and then we can discuss it. For me at that point is totally okay to go with the simplest solution :)

app/src/main/kotlin/com/intive/picover/parties/state/PartiesState.kt
```diff
@@ -5,4 +5,5 @@ import com.intive.picover.parties.model.Party
 sealed class PartiesState {
```
BorysChajdas: @marcinWisnia you will adjust it to use base state from your other merge request after merge?

marcinWisnia: ...or do it later :D but i can adjust, no problem !

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
bartek977: is `runCatching` really needed? wouldn't the `catch` function be enough?
```
partiesRepository.parties()
    .catch { parties.value = Error }
    .collect { parties.value = Loaded(it) }
```

app/src/main/kotlin/com/intive/picover/parties/repository/MockedPartiesRepository.kt
marcinWisnia: ok, used childSnapshot :D

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
marcinWisnia: unified :D (using this in brackets)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: for sure not, especially when we want to use a retry mechanism soon :D i wanted to fix it in separate task for error/retry mechanism impl but i will fix it now :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartyDetailsViewModel.kt
marcinWisnia: for sure not, especially when we want to use a retry mechanism soon :D i wanted to fix it in separate task for error/retry mechanism impl but i will fix it now :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: You are right -> used catch in this way and it works also properly :) 

What is interesting, tests stop working for checking error case, i had to use:

```
every { repository.partyById(partyId) } returns flow<PartyRemote> { throw Exception() }
```
instead of:

```
every { repository.partyById(partyId) } throws Exception()
```

...which convinced me, that i can use parameterised tests, and we can do it for all cases in VM's when we want to check states, as it can be written in this way : 

```
emptyFlow<PartyRemote>() to Loading,
flowOf(partyRemote) to Loaded(party),
flow<PartyRemote> { throw Exception() } to Error,
```
But one doubt come into my mind, in which way we want to do it (as we are using data fetching in init block, and we have to init viewmodel for every single test): 
- In the way i've implemented in this PR for viewmodels, that we have **ONE** test for every value, with forAll usage and initing viewmodel for every iteration, using nice features of kotest forAll, especially in failure info printing
- In the way that is for example implmented in `MainStateTest`, that we have **MANY** tests, with isolation mode per method, without usage of for all, but we don't have to init viewmodel in every single test of parameterised,

You can check this comment @kmaslowiec @BorysChajdas @MarekMacko @bartek977 @michal-kucznerowicz @Nataniel-Antosik  and give an opinion, i don't want to suggest anything now :D

bartek977: I would go with one test ;)

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
Nataniel-Antosik: Same I would go with one test 😄

Nataniel-Antosik (approved):

PR #122, opened 2023-06-12
#64 Use Docker image

### LINKED ISSUE #64: Optimise app builds on CI

Nataniel-Antosik:


.github/workflows/default_branch_verification.yml
```diff
@@ -8,15 +8,12 @@ jobs:
 
   build:
     runs-on: ubuntu-latest
+    container: gradle:8.0.0-jdk17
```
MarekMacko, line 11: Did you check if there is a possibility to setup it in one place, globally?

Nataniel-Antosik: I searched solution to make it globally, but at this moment I can't find the correct way to do it, I will continue to look at it if it is possible

MarekMacko (approved):

marcinWisnia (approved):

PR #128, opened 2023-06-20
#64 Make more readable regex

### LINKED ISSUE #64: Optimise app builds on CI

Nataniel-Antosik:


.github/scripts/branch/checkBranch.sh
```diff
@@ -1,7 +1,15 @@
 #!/bin/bash
 
+acronym="[a-z]{3}"
+slash="\/"
+numbers="[0-9]+"
+hyphen="-"
+word="[^[:cntrl:]]+"
+
+regex="$acronym$slash$numbers$hyphen$word"
```
BorysChajdas: I think that extracting this line to separate variable is redundant, it could be inlined.

Nataniel-Antosik: Done 😄

BorysChajdas (approved):

bartek977 (approved):

PR #129, opened 2023-06-20
#64 Add checking count of commits

### LINKED ISSUE #64: Optimise app builds on CI

Nataniel-Antosik:


.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,14 @@ jobs:
       - name: Check correct commit message
         run: ./.github/scripts/commit/checkCommitMessage.sh
 
+  count-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Count Commits
+        run: ./.github/scripts/commit/checkCountCommits.sh "${{ github.event.pull_request.url }}" "${{ github.repository }}" "${{ secrets.GITHUB_TOKEN }}"
```
MarekMacko: Did you check if there is a direct access to number of commits via actions? Something like `github.event.pullrequest.commits` ?

Nataniel-Antosik: Yes, it's possible to use it therefore I did it and now it looks more easy then before :D

.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,19 @@ jobs:
...
+      - name: PR commits
+        run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} ))" >> "${GITHUB_ENV}"
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+      - name: Count Commits
+        run: ./.github/scripts/commit/checkCountCommits.sh "${{ env.PR_FETCH_DEPTH }}"
+
```
MarekMacko, line 40: Try to call script directly here, seems that flow of passing to script and retrieving the result, only to execute simple assertion looks like an overengineering

Nataniel-Antosik: Yes, I agree with you it looks like overengineering. I changed it 😄

.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,23 @@ jobs:
       - name: Check correct commit message
         run: ./.github/scripts/commit/checkCommitMessage.sh
 
+  count-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: PR commits
+        run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} ))" >> "${GITHUB_ENV}"
```
MarekMacko: Is thins needed as a separate step? Are we able to merge it with `Count Commits` step?

Nataniel-Antosik: I will try it

Nataniel-Antosik: Done 😄

.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,21 @@ jobs:
...
 
+  count-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: ${{ github.event.pull_request.commits }}
+      - name: Count Commits
```
MarekMacko: `Assert commits`

.github/workflows/pull_request_verification.yml
```diff
@@ -20,6 +20,21 @@ jobs:
...
+        run: |
+          if (( ${{ github.event.pull_request.commits }} > 1 )); then
```
MarekMacko: `github.event.pull_request.commits` could be extracted to a variable to not repeat this

.github/workflows/pull_request_verification.yml
MarekMacko: or `Assert number of commits` - to indicate intention of that operation

Nataniel-Antosik: Done

.github/workflows/pull_request_verification.yml
Nataniel-Antosik: Done

bartek977 (approved):

kmaslowiec (approved):

PR #134, opened 2023-06-22
#127 Create common view model state

### LINKED ISSUE #127: Create common view model state

marcinWisnia:


app/src/main/kotlin/com/intive/picover/common/viewmodel/state/ViewModelState.kt
```diff
@@ -0,0 +1,19 @@
+package com.intive.picover.common.viewmodel.state
+
+sealed class ViewModelState<out T> {
```
marcinWisnia, line 3: Was wondering about two things here:

- Adding undefined state
- If we should keep the data within the state -> i can imagine some more complicated view models with more values than one + it could be problematic for Loading Send state

app/src/test/kotlin/com/intive/picover/common/viewmodel/state/ViewModelStateTest.kt
```diff
@@ -0,0 +1,54 @@
...
+					data = null,
+				),
+				ViewModelStateParams(
+					state = ViewModelState.Loaded(data),
+					isLoading = false,
+					isError = true,
+					isLoaded = true,
+					data = data,
+				),
+			).forAll { (viewModelState, isLoading, isError, isLoaded, data) ->
```
marcinWisnia: Kotest gives us `forAll` collection inspectors, which can be used in the tests. 

Output from this is very nice, when the error occurs:

2 elements passed but expected 3

The following elements passed:
ViewModelStateParams(state=com.intive.picover.common.viewmodel.state.ViewModelState$Loading@682a52b8, isLoading=true, isError=false, isLoaded=false, data=null)
ViewModelStateParams(state=com.intive.picover.common.viewmodel.state.ViewModelState$Error@80ecae9, isLoading=false, isError=true, isLoaded=false, data=null)

The following elements failed:
ViewModelStateParams(state=Loaded(data=Any(#1)), isLoading=false, isError=true, isLoaded=true, data=Any(#1)) => The following assertion for Loaded(data=Any(#1)) failed:
expected:<true> but was:<false>
Expected :true
Actual :false

So i think we won't need the dynamic test name for every variant now.
There are also `withData` tests but in my opinion these are quite nice and looks like a candidate for our schema.

MarekMacko: Nice! :)

app/src/main/kotlin/com/intive/picover/common/viewmodel/state/ViewModelState.kt
```diff
@@ -0,0 +1,19 @@
...
+	object Loading : ViewModelState<Nothing>()
+	object Error : ViewModelState<Nothing>()
+	data class Loaded<T>(val data: T?) : ViewModelState<T>()
```
MarekMacko: I don't like this null here. Do we need it or what cases it covers?

marcinWisnia: ah, sorry, a leftover from checking other solutions :D was not my intention to have nullable here -> fixed

app/src/test/kotlin/com/intive/picover/common/viewmodel/state/ViewModelStateTest.kt
```diff
@@ -0,0 +1,54 @@
...
+import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.inspectors.forAll
+import io.kotest.matchers.shouldBe
+import io.mockk.mockk
+
+internal class ViewModelStateTest : ShouldSpec(
+	{
+		val data: Any = mockk()
+
+		should("SHOULD isLoading AND isError AND isLoaded AND data return proper values based on view model state") {
```
MarekMacko: minor: I don't see to point any test with `proper`  in a name, we can actually apply it to every operation because everything should be done in a "proper" way :)

app/src/main/kotlin/com/intive/picover/images/view/ImagesScreen.kt
```diff
@@ -18,27 +18,23 @@ import androidx.compose.ui.Modifier
...
-	when (state) {
-		is Success -> {
-			val uris = state.uris
+private fun ValidateFirebaseStorageState(state: ViewModelState<List<Uri>>) {
+	when {
+		state.isLoaded() -> {
 			LazyColumn(
 				content = {
-					items(uris) {
+					items(state.data()!!) {
```
MarekMacko: Would be nice to have a smart cast here

Nataniel-Antosik:
Please rebase the branch to take a new CI 😄

app/src/test/kotlin/com/intive/picover/common/viewmodel/state/ViewModelStateTest.kt
marcinWisnia: i can agree on that :D fixed

app/src/main/kotlin/com/intive/picover/images/view/ImagesScreen.kt
marcinWisnia: What a nice challenge :D 

Of course I can use `is Loaded` without function `isLoaded()` to have no problems with casting  but i want to follow the DRY rule...

What i found out that works is:

```
@OptIn(ExperimentalContracts::class)
fun <T> ViewModelState<T>.isLoaded(): Boolean {
	contract {
		returns(true) implies (this@isLoaded is Loaded<T>)
	}
	return this is Loaded
}
```
But it looks like overkill for me :D

MarekMacko: Why you think it's an overkill?

marcinWisnia: Mainly because i was not able to create a version of this function that is not an extension function of ViewModelState ( doing similar inside a class was a failure from my site) and it looks weird that we have two functions that are "normal" fuctions, for Loading and Error and this one, as a separate extension outside the class for checking Loaded.

But maybe it is not a big deal and it is better to use smart cast + get rid off data() function and having isLoaded as a separate extension than the current solution :D

MarekMacko: You have to elaborate more, what actually means "_doing similar inside a class was a failure from my site_"? Does the  failure means that it's not possible?

marcinWisnia: yes, i wanted to create `isLoaded` as a class function, similar to `isLoading` and `isError` but could not create a function that works with smart casting without using `<T> ViewModelState<T>.` :/

MarekMacko: I still don't get it... Could you post an example of not working class member function?

marcinWisnia: sure, was trying for example with:

```
@OptIn(ExperimentalContracts::class)
	fun isLoaded(): Boolean {
		contract {
			returns(true) implies (this@ViewModelState is Loaded<T>)
		}
		return this is Loaded
	}
```

MarekMacko: I've checked on my side and finally I see the point :) I'm resolving the issue to see what's others opinion here

app/src/main/kotlin/com/intive/picover/images/view/ImagesScreen.kt
```diff
@@ -47,11 +43,11 @@ private fun ValidateFirebaseStorageState(state: ImagesStorageState) {
 			)
 		}
 
-		is Error -> {
+		state.isError() -> {
 			// TODO-KMA Handle the error
 		}
 
-		is Loading -> PicoverLoader(modifier = Modifier.fillMaxSize())
+		state.isLoading() -> PicoverLoader(modifier = Modifier.fillMaxSize())
```
Nataniel-Antosik, line 48: Probably it could be in new line for a better option for debugging or if it works correctly NVM 😅

app/src/main/kotlin/com/intive/picover/common/viewmodel/state/ViewModelState.kt
```diff
@@ -0,0 +1,19 @@
...
+		this is Loading
+
+	fun isError() =
+		this is Error
+
+	fun isLoaded() =
+		this is Loaded
+
+	fun data() =
+		(this as? Loaded)?.data
```
BorysChajdas: I'm wondering if it wouldn't be better to just force not null here

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/images/view/ImagesScreen.kt
marcinWisnia: NVM :D

app/src/main/kotlin/com/intive/picover/common/viewmodel/state/ViewModelState.kt
marcinWisnia: Yes, i was thinking also about this version :D adjusted this, as i see 2 votes for that solution :D

Nataniel-Antosik (approved):

PR #139, opened 2023-07-03
#121 Show user data

### LINKED ISSUE #121: Profile screen
Requires authentication.

## Basic content:
- avatar
- username
- email

More features could be added up to the developer ideas.

## Future
In the future, this screen could be extended to include user activities such as:
- liked photos
- uploaded photos
- last comments
- bio

## Tasks: 

- [x] Show user data
- [x] Update avatar in Firebase
- [x] Adjust UI add information about edit for avatar
- [x] Update user name
- [x] Handle exception
- [x] Extract same statements from updateAvatar, updateName, fetchProfile in ProfileViewModel 
- [x] Add placeholder for loading
- [x] Validate text

Nataniel-Antosik:
<p float="left" align="center">
  
  
</p>


app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -30,6 +30,15 @@ class AuthRepository @Inject constructor(
 		firebaseAuth.signOut()
 	}
 
+	fun photoUrl() =
```
marcinWisnia: Shouldn't we have one method and one data class that fetch all the user data at once ? I could imagine that we will have 20 values connected with the user, we will probably want to avoid 20 functions:D especially when all the data here is some sort of a user profile data :)

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -30,6 +30,15 @@ class AuthRepository @Inject constructor(
...
+		firebaseAuth.currentUser!!.photoUrl
+
+	fun displayName() =
+		firebaseAuth.currentUser!!.displayName
+
+	fun email() =
+		firebaseAuth.currentUser!!.email
```
marcinWisnia: Shouldn't we use coroutines here ?

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -30,6 +30,15 @@ class AuthRepository @Inject constructor(
 		firebaseAuth.signOut()
 	}
 
+	fun photoUrl() =
+		firebaseAuth.currentUser!!.photoUrl
+
+	fun displayName() =
```
marcinWisnia: Do we have the tests for it :D ?

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -36,3 +69,55 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
 		}
 	}
 }
+
+@Composable
+private fun LabelText(
```
marcinWisnia: I don't think that this needs wrapping as we have only usage of composable Text here, that could be just declared in the main layout code part

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -36,3 +69,55 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
+		modifier = Modifier
+			.fillMaxWidth(),
+		fontSize = 20.sp,
+		fontWeight = FontWeight.Bold,
+		textAlign = TextAlign.Center,
+	)
+}
+
+@Composable
+private fun GreyDivider() {
```
marcinWisnia: Also here

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -16,6 +16,15 @@ class ProfileViewModel @Inject constructor(
 		authRepository.logout()
 	}
 
+	fun userAvatar() =
```
marcinWisnia: Where are tests for this functions ?

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -36,3 +69,55 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
+	Divider(
+		modifier = Modifier
+			.fillMaxWidth()
+			.padding(all = 14.dp),
+		thickness = 1.dp,
+	)
+}
+
+@Composable
+private fun ShowAvatar(
```
marcinWisnia: I think we should avoid such naming in UI composable functions, like with show prefix, that looks to me more like a function triggering something to be shown, or intention to do something for example show some popup that is even not connected strictly with our app. Just Avatar or UserAvatar should be ok, to be more focused on the content itself, not an action.

marcinWisnia (changes requested):

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
BorysChajdas: Yeah, I have the same feeling, we should fetch it all at once and return some profile model 😄

Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
Nataniel-Antosik: Not yet, I will do this

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done 😄

Nataniel-Antosik:
I know tests failed I work on it 😅

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
Nataniel-Antosik: I wrote a test for this ViewModel 😄

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -30,6 +34,14 @@ class AuthRepository @Inject constructor(
 		firebaseAuth.signOut()
 	}
 
+	suspend fun userProfile() = withContext(ioDispatcher) {
+		Profile(
+			photo = firebaseAuth.currentUser?.photoUrl,
```
marcinWisnia: Do we expect, for this particular situation, that the currentUser can be null here ? If not, we can just force not null like for deleting account :) Is there a way to enter this screen not being logged in ? 

Also we can use currentUser only once, using one of kotlin Standard extensions functions to not repeat ourselves in calling it :) 

I am also thinking about the fields, which can really be null ? i am not sure that email can be null for example, when the user is using email ? 

// edit: not null, empty in case of some error :D

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -1,26 +1,87 @@
...
 	val context = LocalContext.current
 	Column(
-		modifier = Modifier.fillMaxSize(),
+		modifier = Modifier
+			.verticalScroll(rememberScrollState())
+			.fillMaxSize()
+			.padding(top = 10.dp),
 		horizontalAlignment = Alignment.CenterHorizontally,
 	) {
+		profile.let {
```
marcinWisnia: i think we can drop let here and just use profile.photo etc in the implementation of layout :) for me it is more readable to have this in this way :)

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
```diff
@@ -25,5 +30,14 @@ class ProfileViewModelTest : ShouldSpec(
 
 			coVerify { authRepository.deleteAccount() }
 		}
+
+		should("profile return Profile data WHEN ProfileViewModelTest was initialized") {
+			runTest(StandardTestDispatcher()) {
+				val profile = Profile(null, "-", "-")
```
marcinWisnia: Do we need a real object here ? we can just use mock as well here :) also please consider if runTest is really needed :)

marcinWisnia (changes requested):

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
marcinWisnia: it was just a question that did not suggest anything .... :D

Nataniel-Antosik: Ok, I know it but when I considered this option I would like to try to use coroutines here :D

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: I'm not attached to it so I will change it :D

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
Nataniel-Antosik: Mockk will be enough but what about runTest() without it and a specific dispatcher this test will fail.

**Edit:** Now I deleted specific dispatcher

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
Nataniel-Antosik: I changed it and I follow the approach with currentUser will be not null and only Profile parameters could be nullable and I used `let` extension scope function :P

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
Nataniel-Antosik: Done

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -30,10 +34,23 @@ class AuthRepository @Inject constructor(
 		firebaseAuth.signOut()
 	}
 
+	suspend fun userProfile() = withContext(ioDispatcher) {
```
BorysChajdas: Not blocking comment :D: Is the IO dispatcher really needed here? I'm just wondering if there was some reason why you added it?

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -78,3 +103,74 @@ private fun DeleteAccountDialog(
 		text = { Text(stringResource(R.string.DeleteAccountConfirmationDialogDescription)) },
 	)
 }
+
+@Composable
+private fun ProfileSegment(
+	state: ViewModelState<Profile>,
+) {
+	if (state.isLoaded()) {
+		UserAvatar(state.data().photo)
```
BorysChajdas: Would be better to call `data()` one time and store it in some variable, or pass it as a parameter to this composable.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -78,3 +103,74 @@ private fun DeleteAccountDialog(
...
+		Divider(
+			modifier = Modifier
+				.fillMaxWidth()
+				.padding(all = 14.dp),
+			thickness = 1.dp,
+		)
+		Text(
+			text = stringResource(id = R.string.UserName),
```
BorysChajdas: Inconsistency: Here you are using named parameters `text = ...` but in the next item you are passing text without using named parameters. Also the `modifier` should be the first parameter.

BorysChajdas (changes requested):

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Ok, I see I forgot to add `text = ...`, and I added a modifier as a second parameter because of it 
image
but I am not attached to it so I can move it to the top. I remember we probably discuss it lately and I probably didn't find any official guide on where it should be. 

I looked at the whole project and I found only one place where the modifier is in the second position 
image
so to make my changes more consistent I will move the modifier to the top, also I don't remember if we discuss it at the guild meeting or if it wasn't necessary😄.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Done 😄

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
Nataniel-Antosik: Back to the comment https://github.com/intive/Picover/pull/139#discussion_r1251129071and considering the current, I think at this moment I didn't need coroutines here and use `viewModelScope.launch` in method `fetchProfile()` in `ProfileViewModel` probably I will add it or change something when I will implement update avatar photo and other profile data. 

So it will look like this: 

`AuthRepository`
```
	fun userProfile() =
		requireUser().let {
			Profile(
				photo = it.photoUrl,
				name = it.displayName!!,
				email = it.email!!,
			)
		}
```
`ProfileViewModel`
```
	private fun fetchProfile() {
		_profile.value = Loaded(authRepository.userProfile())
	}
```
What do you think of it @BorysChajdas, @marcinWisnia

Nataniel-Antosik: Now I changed it

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -61,5 +62,16 @@ class AuthRepositoryTest : ShouldSpec(
 
 			verify { firebaseAuth.currentUser!!.delete() }
 		}
+
+		should("return Profile data WHEN userProfile called") {
+			val profile = Profile(null, "-", "-")
+			every { firebaseAuth.currentUser!! } returns mockk {
+				every { photoUrl } returns null
+				every { displayName } returns "-"
```
BorysChajdas: It is better to return some real values from mocks – this way it is easier to later check how the data could look.

BorysChajdas (approved):

marcinWisnia (approved):

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
```diff
@@ -25,5 +31,12 @@ class ProfileViewModelTest : ShouldSpec(
 
 			coVerify { authRepository.deleteAccount() }
 		}
+
+		should("profile return Profile data WHEN ProfileViewModelTest was initialized") {
+			val profile = mockk<Profile>()
+			coEvery { authRepository.userProfile() } returns profile
```
marcinWisnia: coEvery probably is not needed anymore as we don't have coroutines here :D

Nataniel-Antosik: Yes, it is not needed, I deleted it 😄

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
Nataniel-Antosik: Done 😄

PR #140, opened 2023-07-04
#100 delete account confirmation dialog

BorysChajdas:


app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -31,8 +38,43 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
+				viewModel.onDeleteAccountClick()
+				isDeleteAccountDialogVisible = false
+			},
+			onDismiss = { isDeleteAccountDialogVisible = false },
+		)
+	}
+}
+
+@Composable
+private fun DeleteAccountDialog(
```
Nataniel-Antosik, line 61: Maybe this dialog could be more generic, at this moment I don't know where we can reuse it or if we can be worried about it this moment, maybe when someone would like to add again alert dialog will change it to a more generic 😄

BorysChajdas: Like you said – we can think about making it generic when there will be a need for that 😄

marcinWisnia (approved):

Nataniel-Antosik (approved):

PR #146, opened 2023-07-11
#144 Add authors to README

### LINKED ISSUE #144: Update README
We could add badges to the README.md file more info [here](https://shields.io). Could be discussed, if we want it - if yes - we could choose which ones are needed

- [x] Add badges to README
- [x] Add authors to README

Nataniel-Antosik:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,103 @@
 
 App for documenting memories from Android Guild meetings and parties
 
+## Team:
+
+<table>
+  <tr>
+    <td>
+      <p align="center">
+        <a href="https://github.com/BorysChajdas">
+          <img  src="https://avatars.githubusercontent.com/BorysChajdas" alt="Borys Chajdas" width="70">
+        </a>
+      </p>
+    </td>
+	<td>
+      <p align="center">
+        <a href="https://github.com/MarekMacko">
+          <img src="https://avatars.githubusercontent.com/MarekMacko" alt="Marek Macko" width="70">
+        </a>
+      </p>
+    </td>
+	<td>
+      <p align="center">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,76 @@
-# Picover
+# Picover [![Dev branch verification](https://github.com/intive/Picover/actions/workflows/default_branch_verification.yml/badge.svg)](https://github.com/intive/Picover/actions/workflows/default_branch_verification.yml) [![Pull Request verification](https://github.com/intive/Picover/actions/workflows/pull_request_verification.yml/badge.svg)](https://github.com/intive/Picover/actions/workflows/pull_request_verification.yml)
```
marcinWisnia: i see two commits in this pr :D

README.md
```diff
@@ -1,10 +1,76 @@
...
 
 App for documenting memories from Android Guild meetings and parties
 
+## Team:
+
+<p>
```
marcinWisnia: i am wondering, maybe it will be a way to put more images in one row instead of every row for every contributor ?

README.md
Nataniel-Antosik: Yes, I rebased the last branch which I made the current branch and I forgot to rebase :P

README.md
Nataniel-Antosik: Yes, it is possible I can remove `<p>` and it does look like this: 
image
what do you think about it?

Nataniel-Antosik: Ok, I found a better solution for this 
image

BorysChajdas (approved):

BorysChajdas (approved):

marcinWisnia (approved):

PR #151, opened 2023-07-19
#133 Implement no arg compiler plugin

marcinWisnia:
Alternative for using additional constructor or default values in primary constructor, when a no arg constructor is needed (like in deserializing firebase data) -> feel free to share opinion on if we would like to use it :)

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/common/annotation/NoArgConstructor.kt
+++ b/app/src/main/kotlin/com/intive/picover/common/annotation/NoArgConstructor.kt
@@ -1,0 +1,11 @@
+package com.intive.picover.common.annotation
+
+/**
+ * Needed, when a class has to define a no-argument constructor.
+ * Works configured with [no-arg compiler plugin](https://kotlinlang.org/docs/no-arg-plugin.html).
+ *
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/intive/picover/common/annotation/NoArgConstructor.kt
```diff
@@ -0,0 +1,8 @@
+package com.intive.picover.common.annotation
+
+// Needed, when a class has to define a no-argument constructor.
+// Works configured with no-arg compiler plugin (https://kotlinlang.org/docs/no-arg-plugin.html).
+// Example usage: avoid Firebase DatabaseException, when deserializing object from realtime database to data class.
```
MarekMacko: Is it a comment or doc?

marcinWisnia: ..more like a doc for me :D Adjusted with usage of KDoc syntax :)

BorysChajdas (approved):

bartek977 (approved):

PR #154, opened 2023-07-27
#133 Add error handling in parties and single party screen

marcinWisnia:
error2


app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -27,8 +27,9 @@ class PartiesViewModel @Inject constructor(
 		loadParties()
 	}
 
-	private fun loadParties() {
+	fun loadParties() {
 		viewModelScope.launch {
+			state.value = Loading
```
marcinWisnia, line 32: I am wondering here about the state setting.

Now we are using `init { }` to fetch parties first time, and using loading state at the beginning.
When we want to do a retry, we need to set loading state again, so will be good to have undefined state at the beginning and set loading

But there is a problem with testing, if the state is not a flow and we want to check undefined state, set before init block execution.

I've also realised that we are not using MVI at all, ase we planned to try :D Maybe i can create some proposal of MVI solution here, in other issue #155

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.intive.picover.R
+import com.intive.picover.main.theme.Typography
+
+@Preview
+@Composable
+fun PicoverGenericError(
```
Nataniel-Antosik, line 23: Generic is the correct naming for file, method, etc.? I just wondering 😄

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
```diff
@@ -0,0 +1,51 @@
...
+	message: String = stringResource(id = R.string.GenericErrorMessage),
+	onRetryClick: (() -> Unit)? = null,
```
Nataniel-Antosik, line 25: Maybe it could be just `onClick`?

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
Nataniel-Antosik, line 23: `Generic` in the name is the correct for file, method, etc.? I just wondering 😄

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
Nataniel-Antosik: We were talking about it at the guild meeting and my suggestion during the meeting was: 

if we don't want to add an undefined state we can just add a new method for refreshing data which just invokes another method in VM (yes Marcin we discussed about this method could be redundant 😄), and before invoking this method we can set the state to loading, and thanks to that we can omit doubled loading state.

We can go with doubled loading state exactly like now, or we can go with an undefined state. 

For me whatever you choose, I'm in, because if we rewrite it on MVI probably the problem disappear.

Nataniel-Antosik (changes requested):

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
Nataniel-Antosik: We were talking about it at the guild meeting and my suggestion during the meeting was:

if we don't want to add an undefined state we can just add a new method for refreshing data which just invokes another method in VM (yes Marcin we discussed about this method could be redundant 😄), and before invoking this method we can set the state to loading, and thanks to that we can omit doubled loading state.

We can go with doubled loading state exactly like now, or we can go with an undefined state.

For me whatever you choose, I'm in, because if we rewrite it on MVI probably the problem disappear.

Nataniel-Antosik:
The screen looks correct after device rotation?

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
```diff
@@ -0,0 +1,51 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.intive.picover.R
+import com.intive.picover.main.theme.Typography
+
```
Nataniel-Antosik, line 20: The screen looks correct after device rotation?

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
marcinWisnia: Why not, especially if it's really generic, according to custom click and message :D

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
marcinWisnia: In my opinion, as it is a click dedicated for Retry button click specially, not for all layout or anything else, onRetryClick is proper for this case :D

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/common/error/PicoverGenericError.kt
marcinWisnia: phone_orientation
tablet_orientation

Nataniel-Antosik (approved):

PR #158, opened 2023-08-04
#121 Handle exception

### LINKED ISSUE #121: Profile screen

Nataniel-Antosik:
<p align="center">
 
</p>

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
+++ b/app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
@@ -59,6 +59,10 @@ import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.google.android.gms.oss.licenses.OssLicensesMenuActivity
 import com.intive.picover.R
+import com.intive.picover.common.error.PicoverGenericError
+import com.intive.picover.common.viewmodel.state.ViewModelState.Error
+import com.intive.picover.common.viewmodel.state.ViewModelState.Loaded
+import com.intive.picover.common.viewmodel.state.ViewModelState.Loading
 import com.intive.picover.profile.model.Profile
 import com.intive.picover.profile.viewmodel.ProfileViewModel
 import com.skydoves.landscapist.ImageOptions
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -89,11 +92,18 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
+			is Loaded -> {
+				ProfileSegment(profile.data(), photoLauncher, sheetState, coroutineScope)
+			}
+
+			is Loading -> {
+				CircularProgressIndicator()
+			}
+
+			is Error -> {
+				ErrorProfileSegment(viewModel::fetchProfile)
```
Nataniel-Antosik: @intive/android-guild-szczecin 
Or, if we would like I can go with our generic error placeholder:
<p align="center">
 
</p>

marcinWisnia: i think we could go with our generic error, same for loading and use properly modified PicoverLoader ?

app/src/main/res/values/strings.xml
```diff
@@ -20,4 +20,6 @@
     <string name="GenericErrorMessage">Something went wrong. Please try again later.</string>
     <string name="GenericErrorButtonText">Retry</string>
     <string name="EditUserData">Edit User Data</string>
+    <string name="ProfileError">Something went wrong \nClick to refresh</string>
+    <string name="ProfileErrorButtonRetry">Retry</string>
```
marcinWisnia: I think that we should avoid repeating this generic words. Let's imagine that we will have 10 screens and in every we will have a new word "Retry" added :D i would maybe go more with re-using the GenericErrorButtonText or change the string name for this to RetryButton or something like this :D to rethink :D

marcinWisnia (changes requested):

app/src/main/res/values/strings.xml
Nataniel-Antosik: Yes, I agree with you. I added this string because I thought about the situation when we would like to change something on one screen and don't want to make an impact on the other, but now I will go with the generic one to omit to repeat the same value 😄. Also, I will rename it to `RetryButton` to make the name more meaningful and easier to adjust in other places.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
Nataniel-Antosik: Ok, I will go with a generic solution. What about the loader I have a special view for this in another PR because IMO it is a specific situation with the loader.

app/src/main/res/values/strings.xml
Nataniel-Antosik: Done 😄

marcinWisnia (approved):

kmaslowiec:
There are two commits:
image

BorysChajdas (approved):

Issue #165, opened 2023-09-01
Configure Dependabot


MarekMacko:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates

PR #166, opened 2023-09-01
#165 Add Dependabot configuration

MarekMacko:


.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: gradle
+    directory: "/"
+    schedule:
+      interval: "daily"
```
kmaslowiec, line 6: Would it be better to do it once a week? Daily sounds too bold.  I have real trust issues with that.

MarekMacko: What kind of issues did you encounter?

BorysChajdas (approved):

marcinWisnia (approved):

.github/dependabot.yml
kmaslowiec: We will discuss it during the guild

MarekMacko: Our agreement was to keep everything close to code/PR. Any information about encountered issues is welcome, so don't hesitate to post it here, even if you're not sure if it's valid. That's how we learn from each other 🍻

Issue #167, opened 2023-09-01
Use kotlin directory for kotlin files


MarekMacko:
All kotlin files should be placed under /kotlin directory instead of /java

PR #169, opened 2023-09-01
#167 Drop packaging options

MarekMacko:


app/build.gradle.kts
```diff
@@ -38,11 +38,6 @@ android {
 	composeOptions {
 		kotlinCompilerExtensionVersion = "1.4.7"
 	}
-	packagingOptions {
```
kmaslowiec, line 41: It can be that I wasn't at the guild when it was discussed, but why do we want to include those directories in our APK/AAB?

MarekMacko: I'm wondering what was the purpose of excluding them?

BorysChajdas (approved):

marcinWisnia (approved):

app/build.gradle.kts
kmaslowiec: Maybe less, means better :)

MarekMacko: Fully agree but in terms of redundant configuration, `packagingOptions` in this case. We don't want to struggle with future migrating or maintaining, parts of code, which are not needed

PR #181, opened 2023-09-08
#165 Migrate commit assertion to python

MarekMacko:
The same migration to python as for branch name assertions, along with unit tests


BorysChajdas (approved):

bartek977 (approved):

.github/workflows/pull_request_verification.yml
```diff
@@ -5,27 +5,23 @@ env:
   COMMITS_COUNT: ${{ github.event.pull_request.commits }}
 
 jobs:
-  branch-name:
+  assert-pull-request-metadata:
```
michal-kucznerowicz, line 8: What is the purpose of reducing the number of jobs? Just to be more readable or has it any performance advantages?

michal-kucznerowicz (changes requested):

.github/workflows/pull_request_verification.yml
MarekMacko: The main purpose is to keep it simple. I don't see any benefit of keeping them as two separate jobs, since they're light-weight and fast scripts. Most of the time is consumed by the setup of the environment.

michal-kucznerowicz (approved):

PR #187, opened 2023-09-11
170 if user is not logged in we get NPE

### LINKED ISSUE #170: If user is not logged in we get NPE
[Link to comment](https://github.com/intive/Picover/pull/157#discussion_r1313816378)

kmaslowiec:
#170

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
+++ b/app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
@@ -3,6 +3,7 @@ package com.intive.picover.auth.repository
 import android.net.Uri
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.auth.FirebaseAuthRecentLoginRequiredException
+import com.google.firebase.auth.FirebaseUser
 import com.google.firebase.auth.ktx.userProfileChangeRequest
 import com.google.firebase.storage.StorageReference
 import com.intive.picover.auth.model.AccountDeletionResult
@@ -20,7 +21,7 @@ class AuthRepository @Inject constructor(
 	private val bitmapConverter: BitmapConverter,
 ) {
 
-	private val userAvatarReference = storageReference.child("user/${requireUser().uid}")
+	private val userAvatarReference by lazy { storageReference.child("user/${requireUser().uid}") }
 
 	fun observeEvents() =
 		callbackFlow {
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -80,6 +81,5 @@ class AuthRepository @Inject constructor(
 		return userProfile()
 	}
 
-	private fun requireUser() =
-		firebaseAuth.currentUser!!
+	private fun requireUser(): FirebaseUser = checkNotNull(firebaseAuth.currentUser) { "The Firebase user is null" }
```
BorysChajdas: I don't see a point in changing `!!` to `checkNotNull`. The app will crash in both cases when the user is null but the solution with `!!` is shorter and easier to understand what's is going on.

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -140,16 +140,23 @@ class AuthRepositoryTest : ShouldSpec(
 			val currentUser: FirebaseUser = mockk {
 				every { displayName } returns "Marian Kowalski"
 				every { email } returns userEmail
+				every { uid } returns "id"
```
BorysChajdas: @kmaslowiec here is a reason why your test is not working, you are using "id" instead of "test1234567890" which is used in other place:
```
val storageReference: StorageReference = mockk(relaxed = true) {
			every { child("user/$userUid") } returns referenceToUserAvatar
		}
```
so just change `every { uid } returns "id"` to `every { uid } returns userUid` and test will pass :D

kmaslowiec: > @kmaslowiec here is a reason why your test is not working, you are using "id" instead of "test1234567890" which is used in other place:
> 
> ```
> val storageReference: StorageReference = mockk(relaxed = true) {
> 			every { child("user/$userUid") } returns referenceToUserAvatar
> 		}
> ```
> 
> so just change `every { uid } returns "id"` to `every { uid } returns userUid` and test will pass :D

Yep, I figure that out, but thanks.

BorysChajdas:
@kmaslowiec (for the future) In such cases it's better to update pull request (even change the title) than to close and open new one because we are losing some of the history

kmaslowiec:
> @kmaslowiec (for the future) In such cases it's better to update pull request (even change the title) than to close and open new one because we are losing some of the history

@BorysChajdas That's the funny part. I did not close it, I've just changed the name of the branch it github what automatically delete the PR.

PR #200, opened 2023-09-13
#165 Weaken versions declarations in version catalog

MarekMacko:
Logs from last check https://github.com/intive/Picover/network/updates/721170863 does not contain dependencies with versions defined by `stricly`. There are few possibilities to set a version catalog [check here](https://docs.gradle.org/current/userguide/rich_versions.html#sec:required-version).
After checking the documentation, using `strictly` prevents from creating a PR by dependabot. 
I've used 3 different approaches to check, which one will fit our needs.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -5,10 +5,10 @@ ktlint = "0.50.0"
 lint = "31.0.2"
 
 [libraries]
-android-core = { group = "androidx.core", name = "core-ktx", version.strictly = "1.10.1" }
-android-lifecycle = { group = "androidx.lifecycle", name = "lifecycle-runtime-ktx", version.strictly = "2.6.1" }
+android-core = "androidx.core:core-ktx:1.10.1"
+android-lifecycle = { group = "androidx.lifecycle", name = "lifecycle-runtime-ktx", version.prefer = "2.6.1" }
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -5,10 +5,10 @@
 lint = "31.0.2"
 
 [libraries]
-android-core = { group = "androidx.core", name = "core-ktx", version.strictly = "1.10.1" }
-android-lifecycle = { group = "androidx.lifecycle", name = "lifecycle-runtime-ktx", version.strictly = "2.6.1" }
+android-core = "androidx.core:core-ktx:1.10.1"
```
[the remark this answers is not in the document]
MarekMacko: Lint sarif works! Warnings will be fixed by dependabot later

gradle/libs.versions.toml
```diff
@@ -5,10 +5,10 @@ ktlint = "0.50.0"
...
```
BorysChajdas, line 8: It works? :D Why it is stored like this not using `{ group = ..., name = ..., version... = ...`?

MarekMacko: https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format

image

That's why it's always worth to check documentation 😄

BorysChajdas (approved):

marcinWisnia (approved):

Issue #186, opened 2023-09-11
TextValidator - fix & cleanup 


MarekMacko:
1. Replace `if` with `when` - quick action is available
2. `ProfileViewModel::validatingName` - new instance of `TextValidator.Builder` is created on every recomposition, it shouldn't be even created in view model.

PR #205, opened 2023-09-14
#186 Refactor text validator first step

BorysChajdas:


app/src/main/kotlin/com/intive/picover/validators/TextValidator.kt
```diff
@@ -7,33 +7,26 @@ import com.intive.picover.validators.ValidationStatus.ValidText
...
 	class Builder(private val text: String) {
-		private var allowEmpty: Boolean? = null
-		private var allowBlank: Boolean? = null
+		private var allowEmpty = true
+		private var allowBlank = true
 		private var maxLength: Int? = null
 
 		fun allowEmpty(allow: Boolean) = apply { allowEmpty = allow }
 		fun allowBlank(allow: Boolean) = apply { allowBlank = allow }
 		fun maxLength(length: Int) = apply { maxLength = length }
```
MarekMacko, line 29: As an improvement: this should be also changed to declarative way, e.g.
```
build {
  allowEmpty = true
  allowBlank = false
}

**Review by MarekMacko — APPROVED** [review]

**Comment by MarekMacko** [comment]
Would be nice to name commits in more meaningful name instead of "first step" 😄

File: app/src/main/kotlin/com/intive/picover/validators/TextValidator.kt (hunk 1 above)
Comment (BorysChajdas) [reply app/src/main/kotlin/com/intive/picover/validators/TextValidator.kt:29]: Could you give me a more detailed example because I don't understand? Do you mean to change the Builder to allow creating it like this?: 
```
Textalidator.Builder.build {
    allowEmpty = true
    allowBlank = false
    maxLength = 4
}
```

MarekMacko: Exactly, or it can be function like `validator { <setup here>  }`. Moe inspirations here https://kotlinlang.org/docs/type-safe-builders.html#how-it-works

marcinWisnia (approved):

PR #206, opened 2023-09-14
#110 Add coroutine dispatcher qualifier

### LINKED ISSUE #110: Create a class for handling code and test dispatchers

MarekMacko:


app/src/main/kotlin/com/intive/picover/common/coroutines/dispatcher/Dispatcher.kt
```diff
@@ -0,0 +1,10 @@
+package com.intive.picover.common.coroutines.dispatcher
+
+import javax.inject.Qualifier
+
+annotation class Dispatcher {
+
+	@Retention(AnnotationRetention.BINARY)
```
michal-kucznerowicz: Why do we need to retain this annotation in binary output?

app/src/main/kotlin/com/intive/picover/common/coroutines/dispatcher/Dispatcher.kt
```diff
@@ -0,0 +1,10 @@
...
+	@Qualifier
+	annotation class IO
```
michal-kucznerowicz, line 10: You can also specify annotation target e.g. `@Target(AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.FUNCTION)`

michal-kucznerowicz (changes requested):

app/src/main/kotlin/com/intive/picover/common/coroutines/dispatcher/Dispatcher.kt
MarekMacko: That's the review I was looking for 😄 Good point, I'll change it to `SOURCE`

app/src/main/kotlin/com/intive/picover/common/coroutines/dispatcher/Dispatcher.kt
MarekMacko: I'll set it only to `@Target(AnnotationTarget.FUNCTION, AnnotationTarget.VALUE_PARAMETER)`, `PROPERTY` is not needed at this moment

michal-kucznerowicz (approved):

marcinWisnia (approved):

PR #210, opened 2023-09-15
#186 Inject text validator

BorysChajdas:


marcinWisnia (approved):

app/src/main/kotlin/com/intive/picover/validators/TextValidator.kt
```diff
@@ -6,27 +6,26 @@ import com.intive.picover.validators.ValidationStatus.TooLongText
...
 
 class TextValidator private constructor(
-	private val text: String,
 	private val allowEmpty: Boolean,
 	private val allowBlank: Boolean,
 	private val maxLength: Int?,
 ) {
 
-	fun validatingText() = when {
+	fun validatingText(text: String) = when {
```
MarekMacko, line 14: minor: `validatingText` -> `validate`

app/src/main/kotlin/com/intive/picover/common/validator/qualifier/Validator.kt
```diff
@@ -0,0 +1,10 @@
+package com.intive.picover.common.validator.qualifier
+
+import javax.inject.Qualifier
+
+interface Validator {
+
+	@Qualifier
+	@Retention(AnnotationRetention.RUNTIME)
+	annotation class Profile
```
MarekMacko, line 9: I would for creating a strong type (class) instead on using qualifier since it can be avoided in this case but it's rather a personal approach

MarekMacko (approved):

MarekMacko:
@BorysChajdas Please link the issue to close if after the merge

app/src/main/kotlin/com/intive/picover/validators/TextValidator.kt
BorysChajdas: I know :D. It will be renamed in another pull request, but thanks for catching it!

BorysChajdas:
> @BorysChajdas Please link the issue to close if after the merge

@MarekMacko it is not the last pull request related to refactor of that

app/src/main/kotlin/com/intive/picover/common/validator/qualifier/Validator.kt
BorysChajdas: Do you mean to create a class for each validator? Like `ProfileTextValidator(...) : TextValidator(...)`?

MarekMacko: Exactly, that could be one of the option. 
Apart from that, it shouldn't be named `Profile` since it could suggest that it used for validation of profile fields as a group but it checks only a name, nothing more.

PR #212, opened 2023-09-15
#186 Use DSL builder style for text validator

BorysChajdas:


app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
```diff
@@ -19,13 +19,16 @@ class TextValidator private constructor(
 	}
 
 	class Builder {
-		private var allowEmpty = true
-		private var allowBlank = true
-		private var maxLength: Int? = null
+		var allowEmpty = true
+		var allowBlank = true
+		var maxLength: Int? = null
```
MarekMacko, line 24: getter should be private

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
```diff
@@ -19,13 +19,16 @@ class TextValidator private constructor(
...
 
-		fun allowEmpty(allow: Boolean) = apply { allowEmpty = allow }
-		fun allowBlank(allow: Boolean) = apply { allowBlank = allow }
-		fun maxLength(length: Int) = apply { maxLength = length }
-		fun build() = TextValidator(allowEmpty, allowBlank, maxLength)
+		fun build() =
+			TextValidator(allowEmpty, allowBlank, maxLength)
```
marcinWisnia: minor: Shouldn't this stay as it was :D ?

MarekMacko: Kotlin 1.0 has a rule for that to keep it in one line, as it was before

app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
BorysChajdas: So the property needs to be private and a custom setter needs to be added as it was before. Why in your opinion the getter need to be private?

app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
BorysChajdas: Done

app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
MarekMacko: Oh right, we have a limitation here. 
Why getter should be private? Because this is how the builder design pattern looks like :) 
Every build has only setters and `build` function and accessing e.g. `builder.name` shouldn't be allowed. 
So let's keep it as it is.

MarekMacko (approved):

marcinWisnia (approved):

Issue #208, opened 2023-09-14
Rebuild basic workflow


michal-kucznerowicz:

MarekMacko:
@michal-kucznerowicz I would suggest to extract ktlint check from `pull_request_verification` to a separate workflow file and get rid of redundant steps

Issue #199, opened 2023-09-13
Pass APK between jobs in an app distribution workflow


michal-kucznerowicz:
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#passing-data-between-jobs-in-a-workflow

MarekMacko:
Closed due to current solution that fits our needs

PR #220, opened 2023-09-15
#6 Fix CD workflow

BorysChajdas:


.github/workflows/default_branch_verification.yml
```diff
@@ -6,19 +6,22 @@ on:
...
 jobs:
   build:
+    runs-on: ubuntu-latest
+    container: gradle:8.0.0-jdk17
     steps:
       - name: Build
         uses: ./.github/workflows/setup_build_environment.yml
         with:
           command: ./gradlew build
-        secrets: inherit
```
MarekMacko, line 14: It's not needed anymore?

BorysChajdas: I'm not sure, but it's not allowed when using `steps` that's why I've deleted it. After merging it the develop build should pass because the syntax will be correct but I'm not sure if it will work correctly. I will try to test it on a separate branch later.

MarekMacko: You can just switch temporarily this workflow to trigger on pull request and check if everything will pass

BorysChajdas: It's not passing 😅 changing to draft

michal-kucznerowicz:
It will be fixed in the scope of the https://github.com/intive/Picover/issues/208 by replacing reusable workflow with composite actions. In this step, it won't use artifacts, so https://github.com/intive/Picover/issues/199 can be reopened.

PR #248, opened 2023-09-20
#244 Update status bar color

### LINKED ISSUE #244: Update status bar color
The status bar color should match the app background color based on the theme

michal-kucznerowicz:


app/src/main/kotlin/com/intive/picover/main/theme/StatusBar.kt
```diff
@@ -0,0 +1,14 @@
+package com.intive.picover.main.theme
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
+import androidx.compose.ui.graphics.Color
+import com.google.accompanist.systemuicontroller.rememberSystemUiController
+
+@Composable
+fun SetStatusBarColor(color: Color) {
```
bartek977: as the naming convention for compose doesn't recommend using verbs in composable functions, maybe `ColoredStatusBar`?

michal-kucznerowicz: Done

gradle/libs.versions.toml
```diff
@@ -32,6 +32,7 @@ lint-api = { group = "com.android.tools.lint", name = "lint-api", version.ref =
 ktlint-cli = { group = "com.pinterest.ktlint", name = "ktlint-cli", version.ref = "ktlint" }
 ktlint-ruleset = { group = "com.pinterest.ktlint", name = "ktlint-cli-ruleset-core", version.ref = "ktlint" }
 material = "com.google.android.material:material:1.9.0"
+google-accompanist = "com.google.accompanist:accompanist-systemuicontroller:0.32.0"
```
BorysChajdas: Google-accompanist is not needed to change the color of the status bar. Look at my answer on StackOverflow: https://stackoverflow.com/questions/73756866/jetpack-compose-cant-make-statusbar-completely-transparent/74593754#74593754. It could be achieved without 3rd party library.

app/src/main/kotlin/com/intive/picover/main/theme/ColoredStatusBar.kt
```diff
@@ -0,0 +1,14 @@
+package com.intive.picover.main.theme
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
+import androidx.compose.ui.graphics.Color
+import com.google.accompanist.systemuicontroller.rememberSystemUiController
+
+@Composable
+fun ColoredStatusBar(color: Color) {
```
BorysChajdas: The previous name was "against" composables naming rules but this one is misleading. It suggests that it is a status bar component but in reality, it only executes the side effect. Imo it should not be extracted to separate function but just used directly in `Theme.kt`

BorysChajdas (changes requested):

app/src/main/kotlin/com/intive/picover/main/theme/ColoredStatusBar.kt
MarekMacko: True, probably it won't be reused anywhere

.idea/inspectionProfiles/Project_Default.xml
michal-kucznerowicz, line 1: add to gitignore

gradle/libs.versions.toml
michal-kucznerowicz: Done

BorysChajdas (approved):

bartek977 (approved):

PR #250, opened 2023-09-21
#155 Provide MVI solution for parties screen

### LINKED ISSUE #155: Propose MVI solution for parties screen

marcinWisnia:


app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -0,0 +1,32 @@
+package com.intive.picover.common.viewmodel
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.intive.picover.common.viewmodel.event.Event
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.launch
+
+abstract class MVIViewModel<T : Event> : ViewModel() {
```
marcinWisnia, line 9: Of course this could be done in variant ways:

- Repeat everything in every view model (sucks)
- Have a big abstract class which has everything in it (sucks less :D)
- Use composition (interfaces etc.) 

I see as it as the following:

- Event is the most important thing, without this we don’t have MVI. We also will have repeatable subscribe to events in every event-viewmodel
- If User has only SideEffects, and when they are only UI then, in my opinion, viewmodel is useless :D When there are side effects more like send tracking event to firebase, the viewmodel will be needed
- state can be needed or not - depends on the screen logic and needs

So i’ve prepared events handling in base class, and effects handling in composition way and state just in the viewmodel. Thanks to that You can see different approaches to the topic. We can of course do everything using composition or have events base class + side efects + state as a additional components, it’s worth to discuss.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -19,15 +23,28 @@ import kotlinx.coroutines.launch
 @HiltViewModel
 class PartiesViewModel @Inject constructor(
 	private val partiesRepository: PartiesRepository,
-) : ViewModel() {
+) : MVIViewModel<PartiesEvent>(),
```
marcinWisnia, line 26: Orbit mvi library is doing the following:
- ViewModel implements ContainerHost<State, SideEffect> interface
- ContainerHost have Container val inside
- in viewmodel, we HAVE to override container val, using ViewModel container extension function
- …whichUse viewModelScope.container fun to create the RealContainer object, in which we have all the values like state, effects etc. 
- there are extension functions for ContainerHost like intent, reduce, postSetEffect which handles the container
Looks quite complicated.. :D

app/src/test/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModelTest.kt
```diff
@@ -49,5 +61,25 @@ class PartiesViewModelTest : ShouldSpec(
...
+			clearMocks(repository)
+
+			tested.handleEvents(PartiesEvent.Load)
+
+			coVerify { repository.parties() }
+		}
+
+		should("set navigate to party details side effect WHEN onPartyClick called") {
+			runTest(dispatcher) {
+				val deferredSideEffect = async {
```
marcinWisnia, line 75: if someone knows better way to test shared flow You are welcome :D

PR #251, opened 2023-09-21
#155 Provide MVI solution for parties screen

### LINKED ISSUE #155: Propose MVI solution for parties screen

marcinWisnia:


app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -0,0 +1,32 @@
+package com.intive.picover.common.viewmodel
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.intive.picover.common.viewmodel.event.Event
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.launch
+
+abstract class MVIViewModel<T : Event> : ViewModel() {
```
marcinWisnia, line 9: Of course this could be done in variant ways:

- Repeat everything in every view model (sucks)
- Have a big abstract class which has everything in it(also sucks)
- Use composition (interfaces etc.) 

I see as it as the following:

- Event is the most important thing, without this we don’t have MVI, also every event-VM will have to subscribe to events emitting on init{}
- If User has only SideEffects, and when they are only UI then, in my opinion, viewmodel is useless :D When there are side effects more like send tracking event to firebase, the viewmodel will be needed
- state can be needed or not - depends on the screen logic

So i’ve prepared events handling in base class, and effects handling in composition way and state just in the viewmodel. Thanks to that You can see different approaches to the topic. We can of course do everything using composition or have events base class + side efects + state as a additional components, it’s worth to discuss.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -19,15 +23,28 @@ import kotlinx.coroutines.launch
 @HiltViewModel
 class PartiesViewModel @Inject constructor(
 	private val partiesRepository: PartiesRepository,
-) : ViewModel() {
+) : MVIViewModel<PartiesEvent>(),
+	SideEffectEmitter<PartiesSideEffect> by SideEffectEmitterImplementation() {
```
marcinWisnia, line 27: Orbit mvi library is doing the following:
- ViewModel implements ContainerHost<State, SideEffect> interface
- ContainerHost have Container val inside
- in viewmodel, we HAVE to override container val, using ViewModel container extension function
- …whichUse viewModelScope.container fun to create the RealContainer object, in which we have all the values like state, effects etc. 
- there are extension functions for ContainerHost like intent, reduce, postSideEffect which handles the container

Looks quite complicated.. :D .. but this is done probably also to have viewModelScope in the RealContainer class

app/src/test/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModelTest.kt
```diff
@@ -49,5 +61,25 @@ class PartiesViewModelTest : ShouldSpec(
...
+			clearMocks(repository)
+
+			tested.handleEvents(PartiesEvent.Load)
+
+			coVerify { repository.parties() }
+		}
+
+		should("set navigate to party details side effect WHEN onPartyClick called") {
+			runTest(dispatcher) {
+				val deferredSideEffect = async {
```
marcinWisnia: If someone knows better way to test mutable shared flow You are welcome :D

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+	private val event: MutableSharedFlow<T> = MutableSharedFlow()
+
+	init {
+		subscribeToEvents()
+	}
+
+	open fun handleEvents(event: T) {}
```
bartek977: minor: the name says about events but the param is a single event 😄

marcinWisnia: changed :) thanks !

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -29,30 +30,38 @@ import androidx.navigation.NavHostController
...
 @Composable
 fun PartiesScreen(
 	viewModel: PartiesViewModel,
 	navController: NavHostController,
 ) {
 	val state by viewModel.state
+	LaunchedEffect(true) {
+		viewModel.sideEffect.collectLatest { effect ->
+			when (effect) {
+				is NavigateToPartyDetails -> navController.navigate("partyDetails/${effect.partyId}")
```
MarekMacko, line 48: I'm not convinced that this is the right place. Shouldn't we have some kind of reducer? Where are tests for that?

marcinWisnia: Yeah i know, for now i didn't discover nothing better...I've got only tests for setting the desired side effect.Was thinking also about additional method in the side effects emitter interface that handles the side effects specially for the given VM - then i would probably won't be able to use interface delegation.  But i am not sure how to handle with the nav controller then.

BorysChajdas:
It's a little bit more complicated so I need more time to review and check this. Just leaving a comment to indicate that 😄

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.intive.picover.common.viewmodel.event.Event
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.launch
+
+abstract class MVIViewModel<T : Event> : ViewModel() {
+
+	private val event: MutableSharedFlow<T> = MutableSharedFlow()
```
BorysChajdas: Probably it should be called `events` not `event`.

app/src/main/kotlin/com/intive/picover/common/viewmodel/sideeffect/SideEffectEmitter.kt
```diff
@@ -0,0 +1,10 @@
+package com.intive.picover.common.viewmodel.sideeffect
+
+import kotlinx.coroutines.flow.MutableSharedFlow
+
+interface SideEffectEmitter<T : SideEffect> {
+
+	val sideEffect: MutableSharedFlow<T>
```
BorysChajdas: "sideEffects"

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
BorysChajdas: Imo that's the good starting point. Later on, we could add another layer responsible for navigation. The easiest option for now would be to just store `navController` in some navigator and navigate from there, maybe this way it would be easier to test it as it will be more separated from the view. As far as I know, if we are using a single activity and Compose all the Compose screens have the same `navController` so we only need some kind of "NavControllerHolder" which will store a proper instance of `navController` (attach and detach in activity).

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -19,15 +23,28 @@ import kotlinx.coroutines.launch
...
+
+	fun onPartyClick(partyId: Int) {
+		viewModelScope.launch {
+			setEffect(PartiesSideEffect.NavigateToPartyDetails(partyId))
+		}
+	}
+
+	private fun loadParties() {
 		viewModelScope.launch {
 			state.value = Loading
```
BorysChajdas, line 49: Shouldn't we return the state from the repository? So the `Loading`, `Error` and `Success` with data. If I'm not wrong the data layer should return some kind of state to the ViewModel, and then it could be mapped in ViewModel.

BorysChajdas (changes requested):

BorysChajdas:
Overall for me, that proposition seems promising, and that will be a good starting point to really use MVI :D.

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
marcinWisnia: changed :)

app/src/main/kotlin/com/intive/picover/common/viewmodel/sideeffect/SideEffectEmitter.kt
marcinWisnia: changed :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: hmm, I am not sure about that :D I see this just like in MVVM, that we start with a loading state and then setting error/loaded with data, like i am doing it here :) There is also a way to change the state model itself a little bit, and just, for example, have a separate class for every screen, which has data in it, for example isLoading parameter etc. but i am not sure if it is a good solution :D

bartek977:
let's give it a shot :)

bartek977 (approved):

michal-kucznerowicz (approved):
Looks good to me. Let's give it a try.

kmaslowiec:
> Looks good to me. Let's give it a try.
Same here

kmaslowiec (approved):

BorysChajdas (approved):

PR #255, opened 2023-09-21
#253 Extract dialog

michal-kucznerowicz:


bartek977 (approved):

app/src/main/kotlin/com/intive/picover/profile/view/DeleteAccountDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+		},
+		text = {
+			Text(stringResource(R.string.DeleteAccountConfirmationDialogDescription))
+		},
+	)
+}
+
+@LightDarkPreview
+@Composable
+private fun DeleteAccountDialog() {
```
MarekMacko: Add suffix `Preview`

app/src/main/kotlin/com/intive/picover/profile/view/DeleteAccountDialog.kt
```diff
@@ -0,0 +1,44 @@
...
+	PicoverTheme {
```
MarekMacko, line 41: Is there a way to set this our theme for all previews?

michal-kucznerowicz: I was looking for it but found nothing

app/src/main/kotlin/com/intive/picover/profile/view/DeleteAccountDialog.kt
michal-kucznerowicz: Done

MarekMacko (approved):

PR #259, opened 2023-09-22
#195 avoid tight coupling

### LINKED ISSUE #195: Avoid tight coupling
Internal implementation shouldn't be aware of its outside usage. It should be separated and not coupled, this rule applies for the naming as well. 
Let's take a below example, `openDetails` directly suggest what action should be performed, but it's a responsibility of external component. If we would like to play a video or send tracking event on click, then we would need to change a name to not confuse rest of the team. Trivial thing but can involve to really confusing thing.  
```
@Composable
private fun PartyTile(
    party: Party,
    openDetails: (Int) -> Unit,
) {
    Card(
        modifier = Modifier.clickable {	
            openDetails(party.id)
	},
   ....
```

kmaslowiec:


bartek977 (approved):

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -84,13 +84,13 @@ private fun LoadedContent(parties: List<Party>, openDetails: (Int) -> Unit) {
 @Composable
 private fun PartyTile(
 	party: Party,
-	openDetails: (Int) -> Unit,
+	onPartyClick: (Int) -> Unit,
```
MarekMacko: Here, I would suggest to simplify it to standalone `onClick`

kmaslowiec: Updated. Thank you.

MarekMacko (approved):

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
MarekMacko: You're welcome.

marcinWisnia (approved):

BorysChajdas (approved):

michal-kucznerowicz (approved):

MarekMacko:
@kmaslowiec Is it ready to be merged?

kmaslowiec:
> @kmaslowiec Is it ready to be merged?

Once I update it I will merge it.

PR #262, opened 2023-09-22
#186 Rename validatingText to validate

BorysChajdas:


app/src/test/kotlin/com/intive/picover/validators/TextValidatorTest.kt
```diff
@@ -85,9 +85,9 @@ class TextValidatorTest : ShouldSpec(
 					maxLength = param.maxLength
 				}
 
-				tested.validatingText(param.text) shouldBe param.validationStatus
-				tested.validatingText(param.text).isValid() shouldBe param.isValid
-				tested.validatingText(param.text).errorMessageId shouldBe param.errorMessageResId
+				tested.validate(param.text) shouldBe param.validationStatus
+				tested.validate(param.text).isValid() shouldBe param.isValid
+				tested.validate(param.text).errorMessageId shouldBe param.errorMessageResId
```
MarekMacko, line 90: We should just compare result of `tested.validate(param.text)` instead of calling it three times, would be nice to clean it up

BorysChajdas: Yeah, tests will be refactored in separate PR.

MarekMacko (approved):

michal-kucznerowicz (approved):

PR #263, opened 2023-09-22
#186 Simplify TextValidatorTest

BorysChajdas:


app/src/test/kotlin/com/intive/picover/validators/TextValidatorTest.kt
```diff
@@ -85,20 +70,16 @@ class TextValidatorTest : ShouldSpec(
 					maxLength = param.maxLength
 				}
 
-				tested.validate(param.text) shouldBe param.validationStatus
-				tested.validate(param.text).isValid() shouldBe param.isValid
-				tested.validate(param.text).errorMessageId shouldBe param.errorMessageResId
```
BorysChajdas, line 90: It is not in the scope of `TextValidatorTest` to test the error message or `isValid` function from `ValidationStatus`

MarekMacko: I would even get rid of future test for such properties which just holds id of strings, icons, drawables, etc.

MarekMacko (approved):

kmaslowiec (approved):

PR #268, opened 2023-09-25
#256 User photo error

### LINKED ISSUE #256: User photo error

michal-kucznerowicz:


app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -41,15 +41,14 @@ class AuthRepository @Inject constructor(
...
-	suspend fun userProfile(): Profile {
-		val photoUrl = userAvatarReference.downloadUrl.await()
-		return requireUser().let {
-			Profile(
-				photo = photoUrl,
-				name = it.displayName!!,
-				email = it.email!!,
-			)
-		}
+	fun fetchUserProfile(onSuccess: (profile: Profile) -> Unit, onFailure: () -> Unit) {
```
MarekMacko: You should expose flow, etc. here as a returned type instead using of old-fashioned java callback style

michal-kucznerowicz: I had a solution with `callbackFlow` for all this, but later I started reading that `flow` is unnecessary when we want to get a value once and we don't observe anything so I changed it to what it is

MarekMacko: Could you link the source?

michal-kucznerowicz: I mean the first sentence from flow documentation "In coroutines, a flow is a type that can emit multiple values sequentially, as opposed to suspend functions that return only a single value."

MarekMacko: Understood, then we should rather stick to the previous approach. Was there any particular reason to do that in a callback way?

michal-kucznerowicz: I had to add listeners to Firebase calls to handle it properly to fix this bug. I don't want to create listeners in ViewModel to update the state in its callback to not have Firebase classes in ViewModel. Thus I pass callbacks to the repository this way as I didn't find a better solution for this case.

MarekMacko: Maybe wrapping a suspend function with `kotlin.Result` as a returned type would help?

michal-kucznerowicz: @MarekMacko please take a look at my try with `suspendCoroutine`

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -41,15 +43,14 @@ class AuthRepository @Inject constructor(
...
-	suspend fun userProfile(): Profile {
-		val photoUrl = userAvatarReference.downloadUrl.await()
-		return requireUser().let {
-			Profile(
-				photo = photoUrl,
-				name = it.displayName!!,
-				email = it.email!!,
-			)
-		}
+	suspend fun fetchUserProfile(): Result<Profile> = suspendCoroutine { continuation ->
```
BorysChajdas: Is is needed here to specify the return type?

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -41,15 +43,14 @@ class AuthRepository @Inject constructor(
...
+		userAvatarReference
+			.downloadUrl
+			.addOnSuccessListener {
```
BorysChajdas: Why not just use `await` with `try..catch` instead of adding listeners?

BorysChajdas (changes requested):

app/src/main/kotlin/com/intive/picover/profile/view/UserInfo.kt
```diff
@@ -59,10 +59,7 @@ fun UserInfo(
 				Canvas(
 					modifier = Modifier.size(120.dp),
 					onDraw = {
-						drawRect(
-							brush = brush,
-							size = size,
-						)
+						drawCircle(brush = brush)
```
michal-kucznerowicz, line 62: Fixed shimmer

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
michal-kucznerowicz: Done

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -127,13 +127,21 @@ class AuthRepositoryTest : ShouldSpec(
 		should("return Profile WHEN updateUserAvatar called") {
 			coEvery { referenceToUserAvatar.downloadUrl.await() } returns userPhoto
 
-			tested.updateUserAvatar(userPhoto) shouldBe profile
+			tested
+				.updateUserAvatar(userPhoto)
+				.onSuccess {
+					it shouldBe profile
+				}
```
MarekMacko: In case of exception from `updateUserAvatar` your test will not fail and be still green

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -41,19 +42,18 @@ class AuthRepository @Inject constructor(
...
-				photo = photoUrl,
-				name = it.displayName!!,
-				email = it.email!!,
-			)
+	suspend fun userProfile() =
+		try {
+			val photoUrl = userAvatarReference.downloadUrl.await()
+			Result.success(createProfile(photoUrl))
+		} catch (storageException: StorageException) {
+			Result.success(createProfile())
```
MarekMacko: I would pass explicitly `null` here to `createProfile()` function. Now it's a little bit hidden that we return a profile without uri in case of exception

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
michal-kucznerowicz: Done

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
michal-kucznerowicz: Done and thanks for catching things like this

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
MarekMacko: Not really, I still see the old version

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -127,13 +127,23 @@ class AuthRepositoryTest : ShouldSpec(
...
 		should("return Profile WHEN userProfile called") {
 			coEvery { referenceToUserAvatar.downloadUrl.await() } returns userPhoto
 
-			tested.userProfile() shouldBe profile
+			val result = tested.userProfile()
+
+			result.isSuccess shouldBe true
+			result.onSuccess {
+				it shouldBe profile
+			}
```
MarekMacko: Please check other methods of `kotlin.Result`, there is a nice candidate to combine these two assertions into one :)

MarekMacko (changes requested):

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
michal-kucznerowicz: Thanks, `shouldBeSuccess` will be appropriate

michal-kucznerowicz: Done

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -127,13 +128,16 @@ class AuthRepositoryTest : ShouldSpec(
...
+
+			result shouldBeSuccess profile
 		}
 
 		should("return Profile WHEN userProfile called") {
 			coEvery { referenceToUserAvatar.downloadUrl.await() } returns userPhoto
 
-			tested.userProfile() shouldBe profile
+			val result = tested.userProfile()
+			result shouldBeSuccess profile
```
MarekMacko, line 141: Remember about given-when-then sections

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
MarekMacko: ⬆️

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
MarekMacko: I was thinking about `Result.getOrThrow()` but `shouldBeSuccess` it's much better here. Nice! 🍻

MarekMacko (changes requested):

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
michal-kucznerowicz: Done

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
MarekMacko: Minor: Of course, In such simple cases you can drop inline `result` and keep form: 
```
tested.userProfile() shouldBeSuccess profile
```

MarekMacko (approved):

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
MarekMacko: And I'm also thinking about some lint lint/ktlit rule here to enforce simplifying such cases - but it's a separate topic

marcinWisnia (approved):

app/src/main/kotlin/com/intive/picover/auth/repository/AuthRepository.kt
```diff
@@ -41,19 +42,18 @@ class AuthRepository @Inject constructor(
...
-	suspend fun userProfile(): Profile {
-		val photoUrl = userAvatarReference.downloadUrl.await()
-		return requireUser().let {
-			Profile(
-				photo = photoUrl,
-				name = it.displayName!!,
-				email = it.email!!,
-			)
+	suspend fun userProfile() =
+		try {
```
bartek977, line 46: minor: `runCatching` wraps the response in `Result` out of the box

bartek977 (approved):

BorysChajdas (approved):

PR #270, opened 2023-09-26
#186 Move builder function to companion object

### LINKED ISSUE #186: TextValidator - fix & cleanup

BorysChajdas:
Closes #186

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
+++ b/app/src/main/kotlin/com/intive/picover/common/di/AppModule.kt
@@ -3,8 +3,8 @@ package com.intive.picover.common.di
 import com.google.firebase.database.ktx.database
 import com.google.firebase.ktx.Firebase
 import com.google.firebase.storage.FirebaseStorage
+import com.intive.picover.common.validator.TextValidator
 import com.intive.picover.common.validator.qualifier.Validator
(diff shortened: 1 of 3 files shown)


MarekMacko (approved):
Would also nice to prevent (somehow?) from possibility to call a constructor of `Builder` class outside of the `TextValidator.kt` file

BorysChajdas:
> Would also nice to prevent (somehow?) from possibility to call a constructor of `Builder` class outside of the `TextValidator.kt` file

That would be great but I don't know how to achieve that – making `Builder` a private class is not possible as it is exposed in the public function.

MarekMacko:
> > Would also nice to prevent (somehow?) from possibility to call a constructor of `Builder` class outside of the `TextValidator.kt` file
> 
> That would be great but I don't know how to achieve that – making `Builder` a private class is not possible as it is exposed in the public function.

In the meantime I've found a solution. A hint is context receiver, but let's do it in a separate PR

MarekMacko (approved):

kmaslowiec (approved):

PR #283, opened 2023-09-29
#282 Extract bottom sheet

michal-kucznerowicz:


app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import com.intive.picover.R
+
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalComposeUiApi::class)
+@Composable
+fun ProfileUpdateBottomSheet(
+	userName: String,
+	updateUserName: (String) -> Unit,
+	updateName: (String) -> Unit,
```
michal-kucznerowicz, line 32: Looks strange to me. I will try to merge those functions in the next step.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -0,0 +1,107 @@
...
+						modifier = Modifier.fillMaxWidth(),
+						textAlign = TextAlign.End,
+					)
+				},
+				isError = !validateName(userName),
+				label = { Text(text = stringResource(id = R.string.UserName)) },
+			)
+		}
+	}
+}
```
michal-kucznerowicz, line 107: The preview was not working when I passed this ModalBottomSheet into it. I will try to find a solution and implement it in the next step.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -84,9 +65,7 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
 						)
 					},
 					onEditNameClick = {
-						coroutineScope.launch {
-							sheetState.show()
-						}
+						showProfileUpdateBottomSheet = true
```
MarekMacko, line 55: As an improvement we can think about dedicated delegated property to be used in  way:
- val profileUpdateBottomSheet by bottomSheetVisibility() // or "Presence" etc.
- profileUpdateBottomSheet.show()
- profileUpdateBottomSheet.hide()

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -159,81 +136,16 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
-					label = { Text(stringResource(id = R.string.UserName)) },
-				)
-			}
-		}
+		ProfileUpdateBottomSheet(
+			userName = userName,
+			updateUserName = { userName = it },
+			updateName = { viewModel.updateName(it) },
+			updateVisibility = { showProfileUpdateBottomSheet = it },
+			validateName = { viewModel.validatingName(it).isValid() },
```
MarekMacko, line 112: Of course `viewModel.validatingName(it).isValid()` it shouldn't be called in such way. What is the point of using view model here? 
Just a note, I see that it was used in that way previously

MarekMacko (approved):

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
kmaslowiec: Wouldn't it be the same as using this?
val sheetState = rememberModalBottomSheetState()
val coroutineScope = rememberCoroutineScope()

If we create a class to delegate those booleans with extended ReadWriteProperty I believe we will still need to use rememberSaveable so it would cause a lot of the code and not sure if it is worth the struggle. 

Please see: https://gist.github.com/kmaslowiec/ae22352b61a6fea4a21b79cfe0416819

michal-kucznerowicz: `BottomSheet` visibility and `BottomSheet` state are two different things. If you look at the documentation https://developer.android.com/jetpack/compose/components/bottom-sheets you can see that those properties are used together for different purposes, so I think that we can try to make delegated property for visibility.

MarekMacko: But it's a separate topic, out of the scope of this PR

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
michal-kucznerowicz: As I can see `viewModel` is used only to separate logic. To not use `TextValidator` inside `ProfileUpdateBottomSheet`

MarekMacko: But it doesn't separate too much. `ValidationStatus` is returned from `viewModel::validatingName` but it shouldn't. As previously, it's a separate topic to not block this PR.

marcinWisnia (approved):

bartek977 (approved):

MarekMacko:
@michal-kucznerowicz That should close #282?

michal-kucznerowicz:
> @michal-kucznerowicz That should close #282?

@MarekMacko I want to clean up some logic and add bottom sheet preview in the next PR under #282

PR #285, opened 2023-09-29
#281 Improve options list layout

### LINKED ISSUE #281: Improve options list layout

michal-kucznerowicz:


MarekMacko:
Really nice and neat! 
I think we can use official Github icon

app/src/main/kotlin/com/intive/picover/profile/view/OptionsList.kt
```diff
@@ -0,0 +1,59 @@
...
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import com.google.android.gms.oss.licenses.OssLicensesMenuActivity
+import com.intive.picover.profile.model.OptionsListItem
+
+@Composable
+fun OptionsList(onLogoutClick: () -> Unit, onDeleteAccountCLick: () -> Unit) {
```
MarekMacko: `OptionsList` -> `ProfileActionItem` ? 
`Options` seems to me too generic

app/src/main/kotlin/com/intive/picover/profile/view/OptionsList.kt
```diff
@@ -0,0 +1,59 @@
...
+				},
+				onClick = it.onClick,
+				selected = false,
+			)
+		}
+	}
+}
+
+private object Urls {
+	val github: Uri = Uri.parse("https://github.com/intive/Picover")
```
MarekMacko: 1. `Urls` should be singular
2. Redundant type
3. "https://github.com/intive/Picover" should be placed in resources

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/profile/view/OptionsList.kt
michal-kucznerowicz: 2. "Declaration has type inferred from a platform call, which can lead to unchecked nullability issues. Specify type explicitly as nullable or non-nullable."
3. Do you mean non-translatable string resource?

app/src/main/kotlin/com/intive/picover/profile/view/OptionsList.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/OptionsList.kt
MarekMacko: 2. Ok
3. Yes, it could be a separate resource file, to not mix it with `strings.xml`

michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/model/ProfileActionsListItem.kt
```diff
@@ -0,0 +1,31 @@
+package com.intive.picover.profile.model
+
+import androidx.annotation.DrawableRes
+import androidx.annotation.StringRes
+import com.intive.picover.R
+
+sealed class ProfileActionsListItem(val onClick: () -> Unit, @StringRes val textId: Int, @DrawableRes val iconId: Int) {
```
MarekMacko: `ProfileActionItem` would be enough, there is no need decorate the name too much

app/src/main/kotlin/com/intive/picover/profile/view/ProfileActionsList.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import com.google.android.gms.oss.licenses.OssLicensesMenuActivity
+import com.intive.picover.R
+import com.intive.picover.profile.model.ProfileActionsListItem
+
+@Composable
+fun ProfileActionsList(onLogoutClick: () -> Unit, onDeleteAccountCLick: () -> Unit) {
```
MarekMacko: `ProfileActions` would be enough, there is no need decorate the name too much

app/src/main/res/values/string_constants.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
MarekMacko: `strings_constants` is too generic, it sounds like a bucket for everything. That would lead to a mess in the future. I suggest to keep one file with URLs, for other thing we will create a separate file. 
Suggestion is to rename file to `urls.xml` and drop "Link" from the string name

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/profile/model/ProfileActionsListItem.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/ProfileActionsList.kt
michal-kucznerowicz: Done

app/src/main/res/values/string_constants.xml
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/model/ProfileActionsListItem.kt
MarekMacko: Minor: use singular form "Action", not "Actions"

MarekMacko (approved):

marcinWisnia (approved):

Issue #216, opened 2023-09-15
Provide versionName and versionCode for app distribution


michal-kucznerowicz:
Currently, we can have only one APK uploaded on Firebase. Further uploads don't override it.

MarekMacko:
Change version name to from 1.0 to 0.1 before official first release

PR #300, opened 2023-10-09
#216 Set pre-release version name

MarekMacko:


app/build.gradle.kts
```diff
@@ -1,5 +1,3 @@
-@file:Suppress("DSL_SCOPE_VIOLATION")
```
michal-kucznerowicz, line 1: BTW after the gradle update it can be also removed from `build.gradle.kts`

michal-kucznerowicz (approved):

app/build.gradle.kts
MarekMacko: Thank you for pointing out. I've removed it also from mentioned `build.gradle.kts`

marcinWisnia (approved):

Issue #136, opened 2023-06-29
Articles


Nataniel-Antosik:
## Description for task
We want to add a new screen or just a place where we will add articles to the app. 

## Description to build content
At this moment we just prepare concise/brief presentations for people from Guild as we prepared at the beginning with some cases from private applications or just new topics learned in Android. 

Articles should be written in `Markdown Format`, and should be written `consistently`.

The list to sign up topics for presentations is on the guild channel in the Teams app. 

It is not important to have a finished article when you present a presentation but only a draft with the main points topic.

### Summary
Articles should be:
- written in `Markdown Format`,
- concise/brief.

MarekMacko:
Status: 
- [x] implemented basic screen with data fetching - entry `ArticlesViewModel`,
- [x] dump markdown files are uploaded to Firebase Storage, under `article/`

Todo: 
- [ ] implement basic UI, an inspiration could be a Medium mobile app,
- [ ] display markdown format in composable

PR #304, opened 2023-10-09
#136 Articles basic repository and UI

MarekMacko:


app/src/main/kotlin/com/intive/picover/common/viewmodel/StateFulViewModel.kt
```diff
@@ -0,0 +1,12 @@
...
+import androidx.compose.runtime.State
+import androidx.compose.runtime.mutableStateOf
+import androidx.lifecycle.ViewModel
+import com.intive.picover.common.viewmodel.state.ViewModelState
+
+abstract class StateFulViewModel<T> constructor(initialState: ViewModelState<T> = ViewModelState.Loading) : ViewModel() {
+
+	protected val _state = mutableStateOf(initialState)
+	val state: State<ViewModelState<T>> = _state
+}
```
MarekMacko: Probably all, or almost all, view models will need a state. Instead of repeating the declaration, in every view model, it's reasonable to use basic class here - after a long consideration 😄 
`state` declaration is especially loud and broad in case of List, let's check below example:
```
	val state: MutableState<ViewModelState<List<String>>> = mutableStateOf(ViewModelState.Loading)
```
of course explicit type can be dropped, but we would need to extend `mutableStateOf` with `ViewModelState<List<String>>` type in the following way, to inference proper type:
```
	val state = mutableStateOf<ViewModelState<List<String>>>(ViewModelState.Loading)
```

Using `StateFulViewModel` simplifies this via replacing inheritance from `: ViewModel` to `StateFulViewModel<List<String>>` 
Additionally, introduced backing field (`_state`) would prevent from setting up a state, outside of the view model, without any duplicated declarations. 
That also opens a possibility to create some extensions to handle responses from repositories, like `kotlin.Result` or `Flow`, in a generic way.

app/src/main/kotlin/com/intive/picover/common/viewmodel/StateFulViewModel.kt
```diff
@@ -0,0 +1,13 @@
+package com.intive.picover.common.viewmodel
+
+import androidx.compose.runtime.State
+import androidx.compose.runtime.mutableStateOf
+import androidx.lifecycle.ViewModel
+import com.intive.picover.common.viewmodel.state.ViewModelState
+
+abstract class StateFulViewModel<T> constructor(initialState: ViewModelState<T> = ViewModelState.Loading) : ViewModel() {
```
michal-kucznerowicz: F in StateFul should be lowercase

michal-kucznerowicz (changes requested):

bartek977:
Where are tests? :D

app/src/main/kotlin/com/intive/picover/common/viewmodel/StateFulViewModel.kt
MarekMacko: Done, thanks 🍻

MarekMacko:
> Where are tests? :D

Added 😄

bartek977:
> > Where are tests? :D
> 
> Added 😄

What about viewmodel :D I see you'll extend this screen. so probably the vm also will change, but the init could be tested now ;)

MarekMacko:
> > > Where are tests? :D
> > 
> > 
> > Added 😄
> 
> What about viewmodel :D I see you'll extend this screen. so probably the vm also will change, but the init could be tested now ;)

Added, thank you for pointing out 🍻

bartek977:
One thing, probably we should add test in `ArticlesViewModelTest`  (and others vms which will use `StatefulViewModel`) about initial state, or the `StatefulViewModel` should have appropriate tests.

MarekMacko:
> One thing, probably we should add test in `ArticlesViewModelTest` (and others vms which will use `StatefulViewModel`) about initial state, or the `StatefulViewModel` should have appropriate tests.

Added test for initial state of ArticlesViewModelTest

app/src/test/kotlin/com/intive/picover/articles/viewmodel/ArticlesViewModelTest.kt
```diff
@@ -0,0 +1,37 @@
...
+			tested.state.value shouldBe ViewModelState.Loading
+		}
+
+		should("return set state to error WHEN repository returns failure") {
+			coEvery { articlesRepository.names() } returns Result.failure(Throwable())
+
+			tested.state.value shouldBe ViewModelState.Error
+		}
+
+		should("return set state to loaded WHEN repository returns success") {
```
bartek977, line 30: why return?

bartek977 (approved):

michal-kucznerowicz (approved):

app/src/test/kotlin/com/intive/picover/articles/viewmodel/ArticlesViewModelTest.kt
MarekMacko: Will fix that in upcoming MR ;)

PR #309, opened 2023-10-10
#282 Clean up bottom sheet logic

### LINKED ISSUE #282: Extract bottom sheet

michal-kucznerowicz:


app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
```diff
@@ -26,9 +29,20 @@ class TextValidator private constructor(
...
+
 	class Builder {
 		var allowEmpty = true
 		var allowBlank = true
 		var maxLength: Int? = null
 	}
 }
+
+@Composable
+fun rememberShortTextValidator() = remember {
```
MarekMacko: `remember` doesn't seem to be designed to keep such objects. Created `TextValidator` object is totally stateless, it doesn't keep any value, it's only a processor for passed value so there is no need to remember it between recompositions.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
```diff
@@ -103,14 +103,10 @@ fun ProfileScreen(viewModel: ProfileViewModel) {
...
 	if (showProfileUpdateBottomSheet) {
-		userName = viewModel.userName.value
 		ProfileUpdateBottomSheet(
-			userName = userName,
-			updateUserName = { userName = it },
+			userName = viewModel.userName.value,
 			updateName = { viewModel.updateName(it) },
 			updateVisibility = { showProfileUpdateBottomSheet = it },
-			validateName = { viewModel.validatingName(it).isValid() },
-			getInvalidNameErrorMessageId = { viewModel.validatingName(it).errorMessageId },
```
MarekMacko, line 97: The main issue here is rooted in `ValidationStatus`, for some reason `ValidText` has set `errorMessageId` to `TextIsValid` which is an empty string, in resources. So instead of hiding "error view" we set the text to empty string. Im most of the cases, the view shouldn't visible, not just with no text, because it can occupy some space on the screen due to padding, would that be an expected behavior?

1.I would recommend to set `errorMessageId` for `ValidationStatus` to `null` - one PR

2. view model should keep `TextValidator` and expose only `errorMessageId`. Parameter `validateName` and `getInvalidNameErrorMessageId` can be merged into then.

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -82,24 +97,24 @@ fun ProfileUpdateBottomSheet(
...
+					if (!isUserNameValid) {
 						Text(
 							modifier = Modifier.fillMaxWidth(),
-							text = stringResource(id = getInvalidNameErrorMessageId(userName)),
+							text = stringResource(id = invalidNameErrorMessageId),
 						)
 					}
 					Text(
-						text = "${userName.length} / 20",
+						text = "${userNameText.length} / 20",
```
MarekMacko: Future improvement: Would be nice to connect somehow hardcoded value "20" with `maxLength` from used `TextValidator`. Nobody will remember about keeping these values up to date.

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/common/validator/TextValidator.kt
michal-kucznerowicz: That's true I will keep it in mind

michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/ProfileScreen.kt
michal-kucznerowicz: Done, but the view model doesn't expose errorMessageId. I'm returning it from TextValidator and it is handled by the bottom sheet state

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
```diff
@@ -1,12 +1,12 @@
 package com.intive.picover.common.text
```
michal-kucznerowicz: @kmaslowiec I made some improvements here and fixed the error color. We should reuse this implementation in `ProfileUpdateBottomSheet` in the separate PR.

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
```diff
@@ -24,8 +24,7 @@ fun PicoverOutlinedTextField(
 	keyboardType: KeyboardType = KeyboardType.Text,
 	imeAction: ImeAction = ImeAction.Done,
 	keyboardActions: KeyboardActions = KeyboardActions(),
-	isError: Boolean = false,
-	errorText: String? = null,
+	@StringRes errorMessageId: Int? = null,
```
MarekMacko: It should rather take a `String` as a parameter. We don't want to have two parameters for error, this one and for regular `String`

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -16,25 +16,34 @@ import androidx.compose.material3.ModalBottomSheet
...
-	validateName: (String) -> Boolean,
-	getInvalidNameErrorMessageId: (String) -> Int,
+	onUserNameChange: (String) -> Int?,
 ) {
 	val keyboardController = LocalSoftwareKeyboardController.current
+	var userNameText by remember { mutableStateOf(userName) }
+	var errorMessageId by remember { mutableStateOf(ValidationStatus.ValidText.errorMessageId) }
+	LaunchedEffect(userNameText) {
+		errorMessageId = onUserNameChange(userNameText)
+	}
```
MarekMacko: I have a feeling that it's a little bit messed up. Current value of the name input should be held in view model and each `onValueChange` from `TextField` should update the value in view model. Each change on the name should also trigger a validation and give us an error as a result. In that way we would get rid of used `LaunchedEffect`.

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
marcinWisnia: true, i am also a fan of keeping the editable values in VM, and avoid additional launched effects and remembers :D

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
bartek977: I would also go with `String` parameter

michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -61,7 +59,7 @@ class ProfileViewModel @Inject constructor(
 		}
 	}
 
-	fun updateName(name: String) {
+	fun updateName(name: String = username.value) {
```
bartek977: is the default value needed?

michal-kucznerowicz: It's needed for tests to pass. I'm trying to remove it.

michal-kucznerowicz: I want to change
```
ManageProfileParam(
    { authRepository.updateUserName("Marian K") },
    { tested.updateName("Marian K") },
),
```
to
```
ManageProfileParam(
    { authRepository.updateUserName(any()) },
    { tested.updateName() },
),
```
but I have no idea how to change `profileMethod` in `ManageProfileParam` to be able to pass `any()` inside. I was trying to make it an extension on `MockKMatcherScope` but it doesn't work. Can someone help me with this?

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
```diff
@@ -24,8 +21,7 @@ fun PicoverOutlinedTextField(
 	keyboardType: KeyboardType = KeyboardType.Text,
 	imeAction: ImeAction = ImeAction.Done,
 	keyboardActions: KeyboardActions = KeyboardActions(),
-	isError: Boolean = false,
-	errorText: String? = null,
+	errorMessage: String? = null,
```
MarekMacko: minor: rename to `errorText` ?

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
MarekMacko: @michal-kucznerowicz Please change the PR to draft to notify everyone that it's not ready yet

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
```diff
@@ -65,7 +65,7 @@ class ProfileViewModelTest : ShouldSpec(
 
 			assertSoftly {
 				tested.state.value shouldBe Loaded(profile)
-				tested.userName.value shouldBe "Marian"
+				tested.username.value shouldBe "Marian"
```
MarekMacko: Please use assertSoftly in below way:
```
assertSoftly(tested) {
   state.value shouldBe Loaded(profile)
   username.value shouldBe "Marian"
} 
```

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
MarekMacko: @michal-kucznerowicz Notice that most of the tests from `ProfileViewModelTest` are throwing an exception under the hood, even if they are green. That could be a clue :)

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
michal-kucznerowicz: Done

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
michal-kucznerowicz: Done, default value removed

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -73,15 +71,20 @@ class ProfileViewModel @Inject constructor(
...
 	}
 
-	fun validatingName(text: String) =
-		textValidator.validate(text)
+	fun onUsernameChange(username: String) {
+		_username.value = username
+	}
+
+	private fun validateUsername(text: String) =
+		textValidator.validate(text).errorMessageId
```
MarekMacko: It can be inlined, will be easier to read what's going on in `usernameErrorMessageId`

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -61,11 +60,11 @@ fun ProfileUpdateBottomSheet(
 				actions = {
 					IconButton(
 						onClick = {
-							updateName(userName)
+							updateUsername()
```
MarekMacko: I would improve the name to clearly indicate that we save the data under the hood.
Proposal `saveUserName`

MarekMacko (changes requested):

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
MarekMacko: I meant to get rid of `validateUsername` function and put its body to line 33

michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
```diff
@@ -27,12 +27,11 @@ import com.intive.picover.R
 @OptIn(ExperimentalMaterial3Api::class, ExperimentalComposeUiApi::class)
 @Composable
 fun ProfileUpdateBottomSheet(
-	userName: String,
-	updateUserName: (String) -> Unit,
-	updateName: (String) -> Unit,
+	username: String,
+	saveUsername: () -> Unit,
```
MarekMacko: In my previous comment I was referring to the view model function so:
- this parameter should be changed on `onSaveClick`
- view model function should be named `saveUserName` instead of `updateName`

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -25,14 +27,10 @@ class ProfileViewModel @Inject constructor(
...
-		_state.value.let { state ->
-			if (state is Loaded) {
-				state.data.name
-			} else {
-				""
-			}
-		}
+	private val _username = mutableStateOf("")
+	val username: State<String> = _username
+	val usernameErrorMessageId = derivedStateOf {
```
MarekMacko: Do we actually need `derivedStateOf` here, if each change on `username` triggers recomposition anyway?

app/src/main/kotlin/com/intive/picover/profile/view/ProfileUpdateBottomSheet.kt
michal-kucznerowicz: Done

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
michal-kucznerowicz: Done. There was an unnecessary recomposition trigger. Thanks for catching it 🍻

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -73,15 +69,21 @@ class ProfileViewModel @Inject constructor(
...
 
-	fun validatingName(text: String) =
-		textValidator.validate(text)
+	fun onUsernameChange(username: String) {
+		_username.value = username
+	}
+
+	fun initUsername() {
+		_username.value = state.value.data().name
+	}
```
MarekMacko, line 78: This is something new I guess, why it's needed in `onEditNameClick`?

michal-kucznerowicz: It's needed to init username in the bottom sheet to be the same as showing in the profile so that we can update it. Without this initialization, the text field will have previously typed valued

MarekMacko: So it's seems that we faced first issue with using the same view model for `ProfileScreen` and bottom sheets, which should rather be independent from `ProfileScreen`. That will be more clear when we navigate to bottom sheets by navigation component instead of current "if" approach. Let's keep it as it is but please place a TODO that it needs to be changed. 
It's a nice example of potential cause for future bugs, because entering the bottom sheet for subsequential time depends on the previous one, which should avoid that.

michal-kucznerowicz: Probably the same thing can happen in `PartiesScreen` and its bottom sheet

michal-kucznerowicz: TODO placed

MarekMacko: Seems to be too generic. Rest of the team probably wouldn't what to do. Please place a TODO next to `initUsername` call in `onEditNameClick`. The explanation should contain a clear reason for it, in this case the clue of _"Without this initialization, the text field will have previously typed valued"_

michal-kucznerowicz: Update

MarekMacko: Perfect!

MarekMacko (approved):

marcinWisnia (approved):

app/src/main/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModel.kt
```diff
@@ -25,15 +26,10 @@ class ProfileViewModel @Inject constructor(
...
-				state.data.name
-			} else {
-				""
-			}
-		}
-	}
+	private val _username = mutableStateOf("")
+	val username: State<String> = _username
+	val usernameErrorMessageId
+		get() = textValidator.validate(username.value).errorMessageId
```
bartek977, line 32: if we would use `MutableStateFlow` in viewmodels then we could get rid of such hacks, and use functions like `map`, `combine` which are originally designed for such use cases. I don't block this pr, topic to discuss

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
```diff
@@ -34,7 +34,12 @@ class ProfileViewModelTest : ShouldSpec(
 		val uri: Uri = mockk()
 		val authRepository: AuthRepository = mockk(relaxed = true)
 		val toastPublisher: ToastPublisher = mockk(relaxed = true)
-		val textValidator: TextValidator = mockk()
+		// TODO: Implement separate TextValidator based on screen purpose
+		val textValidator: TextValidator = TextValidator.build {
```
bartek977, line 38: why you use real object here instead of mock?

MarekMacko:
@michal-kucznerowicz Any update here? It blocks #368

BorysChajdas (approved):

app/src/test/kotlin/com/intive/picover/profile/viewmodel/ProfileViewModelTest.kt
michal-kucznerowicz: Mocking every single response for TextValidator it's pointless. In this case for now it's better to use just a real object, but it shows that there is a problem with testing TextValidator in ViewModel so it should be refactored in the next step.

bartek977 (approved):

Issue #194, opened 2023-09-12
Use chained calls from mockk


MarekMacko:
https://mockk.io/#chained-calls

Remove redundant variables by using chained calls.
`listResult` is totally redundant here
```
	val listResult: ListResult = mockk {
		every { items } returns listOf(storageItem)
	}
	every { imageReference.listAll() } returns mockk {
		coEvery { await() } returns listResult
	}
```
Above can be simplified to:
```
	every { imageReference.listAll() } returns mockk {
		coEvery { await().items } returns listOf(storageItem)
	}
```
Goal of this task is to go throughout all test files and simplify then wherever possible

**TODO:**
- [x] ImagesRepositoryTest - use chained calls
- [x] PartiesRepositoryTest - use chained calls
- [x] ImagesViewModelTest - should be parameterized
- [x] #352

michal-kucznerowicz:
Please create subtasks @MarekMacko

MarekMacko:
> Please create subtasks @MarekMacko

Done. @kmaslowiec  and @michal-kucznerowicz I think that would be most beneficial for you to take care of.

kmaslowiec:
> > Please create subtasks @MarekMacko
> 
> Done. @kmaslowiec and @michal-kucznerowicz I think that would be most beneficial for you to take care of.
Ok, I am taking take:
 ImagesRepositoryTest - use chained calls
 PartiesRepositoryTest - use chained calls
 See #284

MarekMacko:
@marcinWisnia @kmaslowiec There are few places in `AuthRepositoryTest` to simplify, please take care of

PR #310, opened 2023-10-11
#194 Parametrize ImagesViewModelTest

michal-kucznerowicz:


app/src/test/kotlin/com/intive/picover/images/viewmodel/ImagesViewModelTest.kt
```diff
@@ -6,40 +6,37 @@ import com.intive.picover.common.viewmodel.state.ViewModelState.Error
...
+				tested.state.value shouldBe state
+			}
 		}
 	},
 )
+
+fun answers(block: MockKStubScope<List<Uri>, List<Uri>>.() -> Unit) =
+	fun MockKStubScope<List<Uri>, List<Uri>>.() {
+		block()
+	}
```
MarekMacko: Really nice, I was trying to find some simplified approach but I don't have any. I think it's worth to use it anyway. Testing against repo response and view model state will be really common, so it's reasonable to invest into it and avoid repeating three boring tests in almost each view model. 
Wrap up: I would go with this solution but let's introduce some small improvements: 
1. Rename function name from `answers` to `mockkAnswer` to avoid name clash with others mockk functions. 
2. Make `mockkAnswer` function as generic to handle other types in different tests
3. Change `tested` to ordinary property, there is no need for `lazy` delegate
4. Move creation of `tested` to "when" block of the test, since our testing part is in `init` block
5. Move `mockkAnswer` to separate file to make it available in other tests files

Later on, we can think about some testing extension to simplify parameter list (lines 25-27) but let's see 

Good job! 🍻

MarekMacko (changes requested):

app/src/test/kotlin/com/intive/picover/images/viewmodel/ImagesViewModelTest.kt
michal-kucznerowicz: I've implemented the proposed improvements

MarekMacko (approved):

PR #317, opened 2023-10-12
#316 Fix large stack of destinations

### LINKED ISSUE #316: Fix large stack of destinations
Pop up to the start destination of the graph to avoid building up a large stack of destinations on the back stack as users select items

michal-kucznerowicz:


kmaslowiec (approved):

kmaslowiec:
https://developer.android.com/jetpack/compose/navigation#bottom-nav

app/src/main/kotlin/com/intive/picover/main/view/MainActivity.kt
```diff
@@ -112,6 +113,7 @@ class MainActivity : ComponentActivity() {
 
 	private fun NavHostController.navigateWithSingleTop(item: NavigationItem) {
 		navigate(item.route) {
+			popUpTo(graph.findStartDestination().id)
```
MarekMacko, line 116: @michal-kucznerowicz Could describe what kind of behavior does it fix? Mentioned issue has only the description copied from the documentation but I'm not sure if anyone knows how it changes the navigation.

michal-kucznerowicz: For example, let's go to PartiesScreen then CameraScreen then PartiesScreen then CameraScreen.
Before this change, the stack looked like: A -> B -> A -> B
After this change, the stack will look like: A -> B
It's just simplified. There is no large stack of destinations.

MarekMacko: Nice, thank you 🍻

Issue #252, opened 2023-09-21
Cleanup ProfileScreen


MarekMacko:
- [x] #253
- [x] #256
- [x] #265
- [x] #281
- [x] #282

PR #325, opened 2023-10-16
#252 Use navigation component for displaying delete account dialog

MarekMacko:


kmaslowiec (approved):

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
```diff
@@ -36,7 +39,17 @@ fun PicoverNavHost(
 			CameraScreen(hiltViewModel())
 		}
 		composable("profile") {
-			ProfileScreen(hiltViewModel())
+			ProfileScreen(hiltViewModel(), navController)
+		}
+		dialog("profile/deleteAccount") {
```
marcinWisnia, line 44: @MarekMacko just curious: Is the profile navigation item still highlighted as selected, on the bottom navigation bar, when the dialog appears ? 

Of course for the dialog it is probably not so important, as we have the rest of the views darkened, but i am wondering if it will work, if we will navigate deeper into something in the one of the bar items screens :D 

Maybe we will should use then some deeper checking of the "group" of screens that belongs to the bar item screens, using `currentDestination.hierarchy` for example. But it's just some bunch of thoughts from my side :D

marcinWisnia (approved):

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
MarekMacko: Good point, it's unselected, so it's the same case as for party details - the "Parties" tab is unselected as well

PR #332, opened 2023-10-18
#329 Drop stateful takenImageUri

MarekMacko:


app/src/main/kotlin/com/intive/picover/camera/viewmodel/CameraViewModel.kt
```diff
@@ -12,13 +11,9 @@ class CameraViewModel @Inject constructor(
 	imageFileRepository: ImageFileRepository,
 ) : ViewModel() {
 
-	val takenImageUri = mutableStateOf<Uri?>(null)
+	val takenImageUri = imageFileRepository.createTempFileAndGetUri()
```
MarekMacko, line 14: Besides weird usage. Please guys remember that not every get/set/create/store operation needs to be placed in **Repository class. Indeed, it's a common pattern for Android apps but you don't have use it every time.

app/src/main/kotlin/com/intive/picover/camera/view/CameraScreen.kt
```diff
@@ -16,13 +16,12 @@ import androidx.compose.ui.res.stringResource
...
 import com.skydoves.landscapist.ImageOptions
 import com.skydoves.landscapist.coil.CoilImage
 
 @Composable
 fun CameraScreen(viewModel: CameraViewModel) {
-	val takenImageUri by viewModel.takenImageUri
+	val takenImageUri = viewModel.takenImageUri
 	val isImageTaken by viewModel.isImageTaken
 	val cameraLauncher = rememberLauncherForActivityResult(ActivityResultContracts.TakePicture()) {
 		viewModel.onImageTaken(wasSaved = it)
```
MarekMacko, line 27: `viewModel.onImageTaken` should be called only, as name says, when image was taken, not for every result from `TakePicture` contract. 
In case of canceled photo, contract returns `false`, which is propagated down to the view model but it shouldn't (parameter `wasSaved` - name is the implication of passed flag) - check `CameraScreen::onImageTaken`

app/src/test/kotlin/com/intive/picover/camera/viewmodel/CameraViewModelTest.kt
MarekMacko, line 1: As a continuation of https://github.com/Android-Guild/Picover/pull/332/files#r1363846875 - note that there is also a missing test for argument `wasSaved = false`

MarekMacko:
Self note: create contract with temp file handling

BorysChajdas (approved):

PR #336, opened 2023-10-19
#329 Enable function naming rule

MarekMacko:


.editorconfig
```diff
@@ -21,6 +21,6 @@ ij_kotlin_imports_layout = *, ^
 ij_kotlin_allow_trailing_comma = true
 ij_kotlin_allow_trailing_comma_on_call_site = true
 ktlint_code_style = android_studio
-ktlint_standard_function-naming = disabled
```
MarekMacko, line 24: It was disabled to lack of possibility to exclude `@Composable` function which starts uppercase

app/src/test/kotlin/com/intive/picover/common/TestUtils.kt
```diff
@@ -3,6 +3,7 @@ package com.intive.picover.common
 import io.mockk.MockKStubScope
 
 fun <T> mockkAnswer(block: MockKStubScope<T, T>.() -> Unit) =
+	@Suppress("ktlint:standard:function-naming")
```
MarekMacko, line 6: @michal-kucznerowicz It's seems to be a ktlint bug, would be nice to create an issue for them

BorysChajdas (approved):

PR #337, opened 2023-10-19
#329 Move creation of temp file for picture to result contract

MarekMacko:


app/src/test/kotlin/com/intive/picover/common/result/ImageFileRepositoryTest.kt
```diff
@@ -1,43 +1,56 @@
...
+import io.kotest.datatest.withData
 import io.kotest.matchers.shouldBe
 import io.mockk.every
 import io.mockk.mockk
+import io.mockk.mockkConstructor
 import io.mockk.mockkStatic
 import io.mockk.unmockkAll
 import java.io.File
 
 class ImageFileRepositoryTest : ShouldSpec(
```
MarekMacko, line 20: To be renamed in next PR to not break the file history

app/build.gradle.kts
```diff
@@ -89,6 +89,7 @@ dependencies {
 	debugImplementation(libs.compose.ui.tooling)
 	debugImplementation(libs.leakcanary)
 	testImplementation(libs.test.kotest.runner)
+	testImplementation(libs.test.kotest.datatest)
```
MarekMacko, line 91: Needed for used `withData` in test

app/src/test/kotlin/com/intive/picover/common/result/ImageFileRepositoryTest.kt
```diff
@@ -1,43 +1,56 @@
...
 			val externalFilesDir: File = mockk()
 			val tempFile: File = mockk()
-			val imageUri: Uri = mockk()
+			val pictureUri: Uri = mockk()
+			val intent: Intent = mockk()
 			every { context.getExternalFilesDir(Environment.DIRECTORY_PICTURES) } returns externalFilesDir
 			every { context.packageName } returns packageName
 			every { File.createTempFile("IMG_", ".jpg", externalFilesDir) } returns tempFile
-			every { FileProvider.getUriForFile(context, "$packageName.provider", tempFile) } returns imageUri
+			every { FileProvider.getUriForFile(context, "$packageName.provider", tempFile) } returns pictureUri
```
MarekMacko, line 43: For future tests we can relax testing approach in cases like this one. We can `any()` instead of exact parameters like `"IMG_"`, ".jpg", etc. because in the end, the test is almost a copy of the implementation, what needs to be tested manually anyway, by launching the app. Such I/O operations should be tested in scope of the instrumentation tests.

marcinWisnia (approved):

Issue #120, opened 2023-06-07
Manage party pictures


BorysChajdas:
On the Party Details screen there should be an option to:
- add a picture from the gallery,
- take a picture and immediately add it.

_____

- There should be an option to delete added images,
- For now, all the party pictures should be displayed in basic form.

Dev notes:
- the image should be compressed to reduce Firebase traffic (it could be extracted to another task),
- the image should be saved to Firebase Storage, and the reference to it should be added to the Firebase database.

MarekMacko:
Dependency #119

MarekMacko:
Status: 
- [ ] staggered grid used for images/photos
- [ ] uploading a picture to firebase storage, scheduling by workmanager

TODO:
- [ ] detail screen for photo
- [ ] delete action

PR #341, opened 2023-10-20
#120 Replace camera screen destination with photos

MarekMacko:


app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
```diff
@@ -35,9 +34,6 @@ fun PicoverNavHost(
 				navController = navController,
 			)
 		}
-		composable("camera") {
-			CameraScreen()
```
MarekMacko, line 39: `CameraScreen` will be merged with `ImagesScreen`

bartek977 (approved):

PR #353, opened 2023-10-24
#352 Simplify chain calls from AuthRepositoryTest

### LINKED ISSUE #194: Use chained calls from mockk

marcinWisnia:


app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
```diff
@@ -137,14 +137,13 @@ class AuthRepositoryTest : ShouldSpec(
```
marcinWisnia, line 133: was also thinking about inlining:

```
val referenceToUserAvatar: StorageReference = mockk {
			every { putFile(userPhoto) } returns mockk {
				every { isComplete } returns true
				every { exception } returns null
				every { isCanceled } returns false
				every { result } returns mockk()
			}
		}
```
with

```
val storageReference: StorageReference = mockk(relaxed = true) {
  every { child("user/$userUid") } returns referenceToUserAvatar
}
```

but i am not sure if referencing to, for example:

` storageReference.child("user/$userUid").downloadUrl.await()`

wouldn't be to complicated in case of getting to know what is going on here :D

bartek977 (approved):

BorysChajdas (approved):

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
MarekMacko, line 133: Please also merge line 94-95, 105-106

MarekMacko (changes requested):
As we discussed, there is no need to create a separate issue for each PR. If there is a parent, like in this one - #194

app/src/test/kotlin/com/intive/picover/auth/repository/AuthRepositoryTest.kt
marcinWisnia: done :)

marcinWisnia:
@MarekMacko You are right with the issue. Should i create a new PR with the correct main issue number  in branch (as i can't edit branch name without new PR) and commit ? 
Or it is ok if i will close the #352 one and leave only #194 as a connected to this PR ?

MarekMacko:
> @MarekMacko You are right with the issue. Should i create a new PR with the correct main issue number in branch (as i can't edit branch name without new PR) and commit ? Or it is ok if i will close the #352 one and leave only #194 as a connected to this PR ?

Let's proceed with the current numer as it is

MarekMacko (approved):

Issue #276, opened 2023-09-28
BottomSheet rotation issue


kmaslowiec:
- [ ] Opened BottomSheet when the rotation is from vertical to horizontal crushes the app and causes: java.lang.IllegalStateException: The offset was read before being initialized. Please see: https://github.com/google/accompanist/issues/1493 

https://github.com/intive/Picover/assets/33199010/d4ba6889-2c2f-4137-9a6c-2c686eda3587

Possible solution: upgrade Material Design to [1.2.0-alpha08]

- [x] The BottomSheet is being hidden during the screen rotation #354 

https://github.com/intive/Picover/assets/33199010/0dc9a1dc-86da-4b37-b40c-3300e4b6194e

michal-kucznerowicz:
Decided to wait for a stable version with a fix.

kmaslowiec:
The issue appears only in our project. Even though the second issue can be fixed with the upgraded Material3, it can be fixed differently. It needs to be addressed.

kmaslowiec:
When we change orientation we are using two different NavHosts. One in PicoverNavigationRail (horizontal) and the other PicoverNavigationBar. 
Hypothesis: Implementing one nav host for both components can resolve the issue.

kmaslowiec:
Maybe rememberUpdatedState? Needs to be investigated.

kmaslowiec:
Source:
Official: https://google.github.io/accompanist/
FAQ: https://medium.com/androiddevelopers/jetpack-compose-accompanist-an-faq-b55117b02712
GitHub: https://github.com/google/accompanist

MarekMacko:
@kmaslowiec Is there anything do to left?

kmaslowiec:
> @kmaslowiec Is there anything do to left?

@MarekMacko Yep, we still need to wait for the stable version that will fix the crash. Please see above: [up](https://github.com/Android-Guild/Picover/issues/276#issue-1917083139)

PR #354, opened 2023-10-25
#276 bottomSheet is hidden after screen rotation

kmaslowiec:
Please see: https://google.github.io/accompanist/navigation-material/
The PR addresses the second issue in #276 

https://github.com/Android-Guild/Picover/assets/33199010/73ea18ab-468d-4453-b060-783396b5d28f

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
+++ b/app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
@@ -9,55 +9,70 @@ import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.dialog
 import androidx.navigation.navArgument
+import com.google.accompanist.navigation.material.BottomSheetNavigator
+import com.google.accompanist.navigation.material.ExperimentalMaterialNavigationApi
+import com.google.accompanist.navigation.material.ModalBottomSheetLayout
+import com.google.accompanist.navigation.material.bottomSheet
 import com.intive.picover.articles.view.ArticlesScreen
 import com.intive.picover.images.view.ImagesScreen
 import com.intive.picover.main.navigation.model.NavigationItem
+import com.intive.picover.parties.view.AddPartyBottomSheet
 import com.intive.picover.parties.view.PartiesScreen
 import com.intive.picover.parties.view.PartyDetailsScreen
 import com.intive.picover.profile.view.DeleteAccountDialog
 import com.intive.picover.profile.view.ProfileScreen
 import com.intive.picover.profile.viewmodel.ProfileViewModel
 
+@OptIn(ExperimentalMaterialNavigationApi::class)
 @Composable
 fun PicoverNavHost(
 	modifier: Modifier,
 	navController: NavHostController,
+	bottomSheetNavigator: BottomSheetNavigator,
 ) {
-	NavHost(
-		modifier = modifier,
-		navController = navController,
-		startDestination = NavigationItem.PARTIES.route,
-	) {
-		composable("parties") {
-			PartiesScreen(
-				viewModel = hiltViewModel(),
-				navController = navController,
-			)
-		}
-		composable("profile") {
-			ProfileScreen(hiltViewModel(), navController)
-		}
-		dialog("profile/deleteAccount") {
-			val viewModel: ProfileViewModel = hiltViewModel()
-			DeleteAccountDialog(
-				onConfirm = {
-					viewModel.onDeleteAccountClick()
-					navController.popBackStack()
-				},
-				onDismiss = { navController.popBackStack() },
-			)
-		}
-		composable("photos") {
-			ImagesScreen(hiltViewModel())
-		}
-		composable(
-			route = "partyDetails/{partyId}",
-			arguments = listOf(navArgument("partyId") { type = NavType.IntType }),
+	ModalBottomSheetLayout(bottomSheetNavigator = bottomSheetNavigator) {
+		NavHost(
+			modifier = modifier,
+			navController = navController,
+			startDestination = NavigationItem.PARTIES.route,
 		) {
-			PartyDetailsScreen(hiltViewModel())
-		}
-		composable("articles") {
-			ArticlesScreen(hiltViewModel())
+			composable("parties") {
+				PartiesScreen(
+					viewModel = hiltViewModel(),
+					navController = navController,
+				)
+			}
+			bottomSheet(route = "parties/addParty") {
+				AddPartyBottomSheet(
+					viewModel = hiltViewModel(),
+					navController = navController,
+				)
+			}
+			composable("profile") {
+				ProfileScreen(hiltViewModel(), navController)
+			}
+			dialog("profile/deleteAccount") {
+				val viewModel: ProfileViewModel = hiltViewModel()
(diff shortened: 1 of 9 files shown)


kmaslowiec:
If we agree to this approach I will create a separate issue to update the bottomSheet in ProfileScreen

gradle/libs.versions.toml
```diff
@@ -11,6 +11,7 @@ android-compose-activity = "androidx.activity:activity-compose:1.8.0"
 android-compose-navigation = "androidx.navigation:navigation-compose:2.7.4"
 android-google-oss-licenses = "com.google.android.gms:play-services-oss-licenses:17.0.1"
 coil = "com.github.skydoves:landscapist-coil:2.2.10"
+compose-accompanist = "com.google.accompanist:accompanist-navigation-material:0.32.0"
```
MarekMacko: Name `compose-accompanist` is too general. Note that there are different artifacts from the same dependency group, e.g. `com.google.accompanist:accompanist-permissions`. In case of including another accompanist dependency we will face the name clash.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -33,28 +29,26 @@ import androidx.navigation.NavHostController
...
-@OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun PartiesScreen(
 	viewModel: PartiesViewModel,
 	navController: NavHostController,
 ) {
 	val state by viewModel.state
-	val sheetState = rememberModalBottomSheetState()
-	val coroutineScope = rememberCoroutineScope()
+
```
MarekMacko: Why it's separated?

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -33,28 +29,26 @@ import androidx.navigation.NavHostController
...
 	LaunchedEffect(true) {
 		viewModel.sideEffects.collectLatest { effect ->
 			when (effect) {
 				is NavigateToPartyDetails -> navController.navigate("partyDetails/${effect.partyId}")
+				is PartiesSideEffect.NavigateToAddParty -> navController.navigate("parties/addParty")
```
MarekMacko: Please do not mix import styles

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
```diff
@@ -9,55 +11,82 @@ import androidx.navigation.compose.NavHost
...
+					viewModel = viewModel,
+					navController = navController,
+				)
+			}
+			bottomSheet(route = "parties/addParty") {
+				val viewModel: PartiesViewModel = hiltViewModel()
+				ModalBottomSheet(
+					onDismissRequest = { navController.popBackStack() },
+				) {
+					PartyBottomSheet(
```
MarekMacko: Name should be changed, it's misleading, there is no bottom sheet under `PartyBottomSheet` composable

MarekMacko:
> If we agree to this approach I will create a separate issue to update the bottomSheet in ProfileScreen

Please address my small suggestions and we're ready to go! That seems to be a right direction with the bottom sheets

gradle/libs.versions.toml
kmaslowiec: done

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
kmaslowiec: Discussed and fixed

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
kmaslowiec: Corrected

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
kmaslowiec: @MarekMacko done

gradle/libs.versions.toml
MarekMacko: accompanist-navigation would be enough but it's up to you

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
MarekMacko: Would be better to keep `PicoverNavHost` clean as possible, thus `ModalBottomSheet` should be embedded to previously named `PartyBottomSheet`, there is no need to keep it separately until there is some reason for that

gradle/libs.versions.toml
kmaslowiec: As in the past was also [Navigation-Animation](https://github.com/google/accompanist/blob/main/navigation-animation) let's keep it that way.

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
kmaslowiec: Updated

MarekMacko (approved):

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
```diff
@@ -9,55 +10,73 @@ import androidx.navigation.compose.NavHost
...
+			navController = navController,
+			startDestination = NavigationItem.PARTIES.route,
 		) {
-			PartyDetailsScreen(hiltViewModel())
-		}
-		composable("articles") {
-			ArticlesScreen(hiltViewModel())
+			composable("parties") {
+				val viewModel: PartiesViewModel = hiltViewModel()
+				PartiesScreen(
```
marcinWisnia, line 40: minor: can we just use hiltViewModel() in constructor of parties screen, without additional value here ?

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
```diff
@@ -9,55 +10,73 @@ import androidx.navigation.compose.NavHost
...
+					viewModel = viewModel,
+					navController = navController,
+				)
+			}
+			bottomSheet(route = "parties/addParty") {
+				val viewModel: PartiesViewModel = hiltViewModel()
+				AddPartyBottomSheet(
+					viewModel = viewModel,
```
marcinWisnia: minor: can we just use hiltViewModel() in constructor of add party bottom sheet, without additional value here ?

marcinWisnia (approved):
@kmaslowiec  nice :)

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
kmaslowiec: done

app/src/main/kotlin/com/intive/picover/main/navigation/view/PicoverNavHost.kt
kmaslowiec: done

PR #370, opened 2023-10-30
#120 Configure and inject SnackbarHostState

MarekMacko:


app/src/main/kotlin/com/intive/picover/images/viewmodel/ImagesViewModel.kt
```diff
@@ -15,6 +16,7 @@ import kotlinx.coroutines.launch
 class ImagesViewModel @Inject constructor(
 	private val repository: ImagesRepository,
 	private val scheduleUploadPhotoUseCase: ScheduleUploadPhotoUseCase,
+	private val snackbarHostState: SnackbarHostState,
```
bartek977, line 19: I have some doubts if we should inject `SnackbarHostState` into vm. I feel the `SnackbarHostState` is like ui component. On the other side, I guess implementation without `SnackbarHostState` in vm can be tricky cuz probably the shared viewmodel will be needed (or passing `SnackbarHostState`) to every screen (what show snakbar). What do you think about it?

marcinWisnia: @bartek977 I see Your point here and mainly agree, but i also think, that maybe this could be a good connection for the side effects. At this moment, showing snackbars and navigating is handled in composable screens. Maybe thanks to this, we will be able to only navigate/show snackbars in the VM, test it well and avoid side effects itself and just do it in VM.

I was also thinking about alternative for this but i don't have a better idea at this moment

MarekMacko: To not go into the details too deeply, `SnackbarHostState` needs be a singleton anyway, so even if we would treat it as a ui component, there is no risk of leaks after configuration change, process death etc. 
Of course, we can add fancy facade, but let's keep it simple until nobody see a reason to do it in a different way :)

app/src/main/kotlin/com/intive/picover/images/viewmodel/ImagesViewModel.kt
```diff
@@ -24,14 +26,22 @@ class ImagesViewModel @Inject constructor(
...
 
 	fun scheduleUploadPhoto(uri: Uri) {
-		scheduleUploadPhotoUseCase(uri)
+		viewModelScope.launch {
+			runCatching {
+				scheduleUploadPhotoUseCase(uri)
+			}.onSuccess {
+				snackbarHostState.showSnackbar("Upload scheduled successfully")
+			}.onFailure {
+				snackbarHostState.showSnackbar("Error: $it")
```
bartek977, line 35: hardcoded strings

MarekMacko: Yes, it's intended to keep it in this way for the initial iteration

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/images/viewmodel/ImagesViewModel.kt
MarekMacko: @bartek977 @marcinWisnia Any conclusion here?

app/build.gradle.kts
```diff
@@ -94,7 +94,7 @@ dependencies {
 	testImplementation(libs.test.guava)
 	testImplementation(libs.test.kotest.datatest)
 	testImplementation(libs.test.mockk)
-	testCompileOnly("org.jetbrains.kotlin:kotlin-reflect:1.9.20") {
+	testCompileOnly("org.jetbrains.kotlin:kotlin-reflect:1.9.10") {
```
MarekMacko, line 97: Fixes:
```
Could not resolve org.jetbrains.kotlin:kotlin-reflect:1.9.20.
     Required by:
         project :app
      > Cannot find a version of 'org.jetbrains.kotlin:kotlin-reflect' that satisfies the version constraints:
           Dependency path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:1.9.20' because of the following reason: Needed to locally trigger single kotest test - check new versions of kotlin and kotest plugins to fix this workaround
           Constraint path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:***strictly 1.9.10***' because of the following reason: debugUnitTestRuntimeClasspath uses version 1.9.10

   > Could not resolve org.jetbrains.kotlin:kotlin-reflect:***strictly 1.9.10***.
     Required by:
         project :app
      > Cannot find a version of 'org.jetbrains.kotlin:kotlin-reflect' that satisfies the version constraints:
           Dependency path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:1.9.20' because of the following reason: Needed to locally trigger single kotest test - check new versions of kotlin and kotest plugins to fix this workaround
           Constraint path 'Picover:app:unspecified' --> 'org.jetbrains.kotlin:kotlin-reflect:***strictly 1.9.10***' because of the following reason: debugUnitTestRuntimeClasspath uses version 1.9.10

PR #372, opened 2023-10-30
#120 Drop acronym from branch template

MarekMacko:


README.md
```diff
@@ -18,12 +18,11 @@ Current issues are present in the [project backlog](https://github.com/orgs/inti
 ### Branch
 
 Each branch should be associated with a team member and corresponding issue created [here](https://github.com/intive/Picover/issues).
-To accomplish that, use following template `acronym/issue_number/description`, example:
-> kma/1/initial-project
+To accomplish that, use following template `issue_number/description`, example:
+> 1/initial-project
```
MarekMacko, line 22: Let me know what do you think

bartek977: I use alike template in my project and I can say the schema with acronym helps with finding a branch with given author.

marcinWisnia: on the other hand, we can see the branch author in the git history as well :D i am 60 to 40 to go with the new approach :D

MarekMacko: @bartek977 Agree, that can be helpful but I would say that it's really uncommon, additionally taking into consideration that usually we don't have more than 5 feature branches so it's not a big deal anyway.

marcinWisnia (approved):

README.md
bartek977: @marcinWisnia I meant the acronym is helpful if you checkout on the branch using android studio plugin
image


I'm not blocking this change, for me both schemas are ok

BorysChajdas (approved):

michal-kucznerowicz (approved):

README.md
kmaslowiec: I personally like that it gives the some kind of the order in my local git:
image
But it is just my personal preference so up to you guys.

PR #398, opened 2023-11-08
#120 Define global failure image

MarekMacko:


app/src/main/kotlin/com/intive/picover/main/theme/Theme.kt
```diff
@@ -54,9 +60,14 @@ fun PicoverTheme(
 			}
 		}
 	}
-	MaterialTheme(
-		colorScheme = colorScheme,
-		typography = Typography,
-		content = content,
-	)
+	val imageComponent = imageComponent {
+		+PlaceholderPlugin.Failure(painterResource(R.drawable.ic_avatar_placeholder))
```
BorysChajdas, line 64: Could it be also used for loading state?

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/main/theme/Theme.kt
MarekMacko: Yes, it can - https://github.com/skydoves/landscapist#placeholderplugin

PR #400, opened 2023-11-09
#120 Define global shimmer effect

MarekMacko:


app/src/main/kotlin/com/intive/picover/images/view/ImagesScreen.kt
```diff
@@ -49,8 +52,13 @@ private fun PhotosGrid(uris: List<Uri>, onImageTaken: (Uri) -> Unit) {
 			verticalItemSpacing = 4.dp,
 			horizontalArrangement = Arrangement.spacedBy(4.dp),
 		) {
-			items(uris) {
-				Photo(uri = it)
+			items(photos) {
+				CoilImage(
+					modifier = Modifier.size(it.width.dp, it.height.dp),
```
MarekMacko, line 57: We need to know the size in advance to prevent from "jumping" the views. For now it's generated in `Photo`, in the  end it will be retrieved from API as a metadata

bartek977: or we can get the resolution from `ExifInterface` :)

BorysChajdas (approved):

bartek977 (approved):

PR #412, opened 2023-11-13
#319 Pass TextValidator to Text composable

### LINKED ISSUE #319: Max length of signs counter depends on TextValidator

MarekMacko:
Closes #319


app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
```diff
@@ -1,71 +1,97 @@
...
 	onValueChange: (String) -> Unit,
 	maxLines: Int = 1,
 	keyboardType: KeyboardType = KeyboardType.Text,
 	imeAction: ImeAction = ImeAction.Done,
 	keyboardActions: KeyboardActions = KeyboardActions(),
-	errorText: String? = null,
 	isEnabled: Boolean = true,
+	validator: TextValidator,
+	isValid: (Boolean) -> Unit = {},
 	trailingIcon: @Composable (() -> Unit)? = null,
```
MarekMacko, line 39: @kmaslowiec It's not set anywhere since the creation of that component, what was the origin? Looks like copy-paste

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
```diff
@@ -1,71 +1,97 @@
...
+	labelText: String,
 	onValueChange: (String) -> Unit,
 	maxLines: Int = 1,
 	keyboardType: KeyboardType = KeyboardType.Text,
 	imeAction: ImeAction = ImeAction.Done,
 	keyboardActions: KeyboardActions = KeyboardActions(),
-	errorText: String? = null,
 	isEnabled: Boolean = true,
+	validator: TextValidator,
+	isValid: (Boolean) -> Unit = {},
```
bartek977: `isValid` suggests the function will check if the value is valid, maybe `onValidationStatusChange` ?

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
kmaslowiec: @MarekMacko isEnabled parameter was added just in case, same as trailingIcon. It would likely be used shortly. IMO we can keep it. If that is not your question, please specify.

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
MarekMacko: Changed to `onValidationStatusChange`, but personally I would keep `isValid` 😃

app/src/main/kotlin/com/intive/picover/common/text/PicoverOutlinedTextField.kt
MarekMacko: Thanks for the explanation. I would suggest to not add the code if you're not sure if it will ever be used. Keep it simple ;)

PR #413, opened 2023-11-13
#155 Simplify MVI side effects

MarekMacko:


app/src/main/kotlin/com/intive/picover/common/viewmodel/sideeffect/SideEffect.kt
```diff
@@ -1,3 +0,0 @@
-package com.intive.picover.common.viewmodel.sideeffect
-
-interface SideEffect
```
MarekMacko: I don't see any point to keep interface for it, simple `Any` seems to be enough

app/src/main/kotlin/com/intive/picover/common/viewmodel/sideeffect/SideEffectEmitterImplementation.kt
```diff
@@ -1,12 +0,0 @@
-package com.intive.picover.common.viewmodel.sideeffect
-
-import kotlinx.coroutines.flow.MutableSharedFlow
-
-class SideEffectEmitterImplementation<T : SideEffect> : SideEffectEmitter<T> {
```
MarekMacko, line 5: Merged with `SideEffectEmitter` and moved to `MVIViewModel`

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -22,8 +20,7 @@ class PartiesViewModel @Inject constructor(
 	private val partiesRepository: PartiesRepository,
 	@Validator.ShortText private val shortTextValidator: TextValidator,
 	@Validator.LongText private val longTextValidator: TextValidator,
-) : MVIViewModel<PartiesState, PartiesEvent>(initialState = PartiesState()),
-	SideEffectEmitter<PartiesSideEffect> by SideEffectEmitterImplementation() {
```
MarekMacko, line 22: Delegation dropped and moved to `MVIViewModel`

app/src/test/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModelTest.kt
```diff
@@ -68,27 +61,11 @@ class PartiesViewModelTest : ShouldSpec(
 		}
 
 		should("start loading parties WHEN load parties event is emitted") {
-			tested.handleEvent(PartiesEvent.Load)
+			tested.emitEvent(PartiesEvent.Load)
```
MarekMacko, line 58: `handleEvent` should be used only within view model, visibility has been changed to `protected`

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -56,8 +56,8 @@ fun PartiesScreen(
 		MVIState.Type.LOADING -> PicoverLoader(modifier = Modifier.fillMaxSize())
 		MVIState.Type.LOADED -> LoadedContent(
 			parties = state.parties,
-			onPartyClick = viewModel::onPartyClick,
-			onFabClick = viewModel::onAddPartyClick,
```
MarekMacko, line 60: That's the overhead of architecture, note how many implication it causes. Check the implementation of, the same goes for unit tests of these functions, it's just redundant.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -56,8 +56,8 @@ fun PartiesScreen(
...
+			onPartyClick = { viewModel.emitEffect(NavigateToPartyDetails(it)) },
+			onFabClick = { viewModel.emitEffect(NavigateToAddParty) },
```
MarekMacko, line 60: Please check where is the handling of these effects - few lines above, in line 50. We emit some side effect through the `flow`, what involves coroutines and make tests more complex, just to handle it few lines above. What is the benefit of that? Separation? Coverage by unit tests? 
Note that we tested only the "facade", `navContoller.navigate(...)` was not touched, what is the most important part here. So we end up with unit tests of intermediate steps and flow but the most important part is missing. That is really nice lesson, we should be more careful about writing unit tests and think about gained value from them. 
It's not the way that we should test everything because "it's a good practice to good to have tests blah blah". 
Unit test are for preventing from the regression, documenting the code and making the code more stable not to increase the statistics of covered lines.

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -10,34 +10,39 @@ import com.intive.picover.common.viewmodel.state.MVIState
...
-	open fun handleEvent(event: E) {}
+	protected abstract fun handleEvent(event: E)
 
 	fun emitEvent(newEvent: E) {
 		viewModelScope.launch {
 			events.emit(newEvent)
 		}
 	}
 
+	fun emitEffect(sideEffect: SideEffect) {
```
MarekMacko: Renamed from `setEffect`

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -56,8 +56,8 @@ fun PartiesScreen(
...
 		)
 
 		MVIState.Type.ERROR -> PicoverGenericError(onRetryClick = { viewModel.emitEvent(PartiesEvent.Load) })
```
MarekMacko, line 63: Comment to restricted visibility of `MVIViewModel::handleEvent`
Please always remember about visibility of functions and properties in basic classes, we shouldn't expose too much. It prevents from unintended usage of our class and makes more clear how it should be used. 
In this case, we could call `viewModel.handleEvent(PartiesEvent.Load)`, it would be even faster and more efficient :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
```diff
@@ -22,8 +20,7 @@ class PartiesViewModel @Inject constructor(
...
+) : MVIViewModel<PartiesState, PartiesEvent, PartiesSideEffect>(initialState = PartiesState()) {
 
 	init {
 		loadParties()
```
MarekMacko, line 22: Data should be loaded according to sent intents/events, according to MVI, so it should be changed. Would be nice if someone would take care of that. 
Please do not create additional issue for that, just create a PR and reference #155 issue

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
```diff
@@ -10,34 +10,39 @@ import com.intive.picover.common.viewmodel.state.MVIState
...
 import kotlinx.coroutines.launch
 
-abstract class MVIViewModel<S : MVIState, E : Event>(initialState: S) : ViewModel() {
+abstract class MVIViewModel<S : MVIState, E : Event, SideEffect : Any>(initialState: S) : ViewModel() {
 
 	@Suppress("ktlint:standard:property-naming")
 	protected val _state = mutableStateOf(initialState)
 	val state: State<S> = _state
-	private val events: MutableSharedFlow<E> = MutableSharedFlow()
+	val sideEffects = MutableSharedFlow<SideEffect>()
```
bartek977: as we already have private mutable field for `state` maybe it's good idea to do the same for `sideEffects` as it should only change in vm?

MarekMacko: Good point, I'll improve that when rest of the feedback will be posted.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
bartek977, line 60: probably we could get rid of all `onXXXClick` as they use the same method (viewModel.emitEffect), and have only one like `onEvent = viewModel::emitEffect`

MarekMacko: It's done in this way or missed something?

bartek977: I mean we have 
```
onPartyClick = { viewModel.emitEffect(NavigateToPartyDetails(it)) },
onFabClick = { viewModel.emitEffect(NavigateToAddParty) },
```
but it could be
```
onEvent = viewModel::emitEffect
```

like this:
```
private fun LoadedContent(
	parties: List<Party>,
        onEvent: (PartiesEvent) -> Unit
) {
```

and then in in onClicks use the onEvent with appropriate event

MarekMacko: Imagine case when `PartiesEvent` will be extended with additional events. Without any warning, it will allow to call `onEvent` callback with events that are not related to `LoadedContent` composable. Side effects are too general to pass them down in this case. It's the same reason why don't pass the view model down to each composable.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
BorysChajdas: How do you see it? Emit some event from view at the beginning to initialize the VM? Could you elaborate more because for me loading something in the constructor is much simpler than sending event to vw to load initial data on screen load.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
```diff
@@ -56,8 +56,8 @@ fun PartiesScreen(
 		MVIState.Type.LOADING -> PicoverLoader(modifier = Modifier.fillMaxSize())
 		MVIState.Type.LOADED -> LoadedContent(
 			parties = state.parties,
-			onPartyClick = viewModel::onPartyClick,
-			onFabClick = viewModel::onAddPartyClick,
+			onPartyClick = { viewModel.emitEffect(NavigateToPartyDetails(it)) },
```
BorysChajdas, line 59: I'm thinking about it and in that case the VM is totally redundant part of the navigation, we are calling `emitEffect` to just react to that effect and navigate. We could directly call `onPartClick =  { navController.navigate("partyDetails/${effect.partyId}") }`, because VM is not taking any part in the side-effect handling. Maybe it's only like that in the simplest cases like this, I'm just loudly thinking. And of course that is not related to your changes because it was even more complicated in the previous approach.

BorysChajdas (approved):

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
bartek977: Agree with @BorysChajdas , I got you @MarekMacko it will be concise with MVI, but I think the patterns are for developers and we can be agile and not stick to the rules always

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
bartek977: I think passing the viewmodel is other case, we don't do it (or we shouldn't) cuz the tests.

I see your point, but to be honest, as it's only for us, we don't share the api for others, so the chance someone will call onEvent with wrong Event shouldn't happen.

Nevertheless it's only idea, I'm not forcing it :)

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
marcinWisnia: Good point @BorysChajdas. I think that probably the best idea will be to use emitEffect here when there will be some sort of a tested navigator in VM. At this moment for this particular situation i have the same feelings.

On the other hand, i am not sure how to do it properly and clean, so the solution with emitting effects and navigation in screen is also ok for me.

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: ...so finally, after some time, we decided to have one abstract class for events, state and side effects :D just like in my guild presentation code :D

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
bartek977: I agree in this case the vm is redundant, but if we go with your idea then the ui will keep part of the logic (what should happen when a user clicks on something).

Edited: Okey, now I see the ui already know about what happen as the event is called `NavigateToPartyDetails`  😄 So we probably could go with your idea, or refactor it, to tell vm user clicks on something and the vm should decide what to do

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
marcinWisnia: @MarekMacko  do You mean here, to emit Load event on init instead of loadParties() usage ?

marcinWisnia (approved):

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
bartek977: @marcinWisnia it would be the easiest, but it would break the pattern as the events should be emitted by view

michal-kucznerowicz (approved):

app/src/main/kotlin/com/intive/picover/common/viewmodel/MVIViewModel.kt
MarekMacko: Done

app/src/main/kotlin/com/intive/picover/common/viewmodel/sideeffect/SideEffect.kt
MarekMacko, line 1: I've reverted the change. In the end I decided to keep it. Basic type/interface helps understand the relation between components and even suggest what name should be used.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
MarekMacko: I would suggest to create some sort of "processor" to handle navigation events. Note that current handling of navigation from side effects is done in `Composable` scope. Now let's think about testing, you have to engage compose framework to execute such test, but `navController::navigate` call is not related to, so there is a high coupling between these to - it should be removed.

app/src/main/kotlin/com/intive/picover/parties/view/PartiesScreen.kt
MarekMacko: I see your point as well and thank you for sharing it :) 

> as it's only for us, we don't share the api for others

In long term, such approach leads to hacky and unmaintainable code :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
MarekMacko: So some state is set by constructor and some by event? Note that creation, and calling the constructor, is determined by the view. In the same way we can handle some clicks by `viewModel::onClick` and some by sending events. 
How would you like to handle call rest of the lifecycle events then (onStop, back clicks, etc.) ?
I think that everyone can predict what would happen with the code after few months :)

app/src/main/kotlin/com/intive/picover/parties/viewmodel/PartiesViewModel.kt
MarekMacko: Probably that was inevitable 😄

Issue #119, opened 2023-06-07
Manage new party


BorysChajdas:
**Add/Edit/Delete**
- [x] Create a screen to add, should be the [bottom sheet](https://m3.material.io/components/bottom-sheets/overview) screen should be entered from the "Parties" screen using **floating action button** #320 #271 #424 
- [ ] We can save the new party using the bottom sheet. The party appears on the list.
- [ ] There should be a possibility to delete and edit the party.

kmaslowiec:
https://medium.com/@manojbhadane/mastering-android-jetpack-compose-bottomsheet-with-material-3-e61af75c0cac

MarekMacko:
@kmaslowiec Is it really finished?

kmaslowiec:
> @kmaslowiec Is it really finished?

Definitely not

PR #424, opened 2023-11-28
#119 Repository to save party

kmaslowiec:
Continuation of https://github.com/Android-Guild/Picover/pull/387

### DIFF
--- a/app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
+++ b/app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
@@ -5,6 +5,7 @@ import com.google.firebase.database.snapshots
 import com.intive.picover.parties.model.PartyRemote
 import javax.inject.Inject
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.tasks.await
 
 class PartiesRepository @Inject constructor(
 	firebaseDatabase: FirebaseDatabase,
@@ -17,6 +18,15 @@ class PartiesRepository @Inject constructor(
 		partiesReference.snapshots.map { snapshot ->
 			snapshot.children.map { childSnapshot ->
 				childSnapshot.getValue(PartyRemote::class.java)!!
+			}
+		}
+
+	suspend fun addParty(partyRemote: PartyRemote) =
+		runCatching {
+			partiesReference.push().let { databaseReference ->
+				databaseReference.key?.let { key ->
+					databaseReference.setValue(partyRemote.copy(id = key)).await()
+				} ?: throw Exception("Firebase returns key as null")
 			}
 		}
 
--- a/app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
+++ b/app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
@@ -5,28 +5,40 @@ import com.google.firebase.database.DatabaseReference
 import com.google.firebase.database.FirebaseDatabase
 import com.google.firebase.database.Query
 import com.google.firebase.database.snapshots
+import com.intive.picover.common.coroutines.CoroutineTestExtension
 import com.intive.picover.parties.model.PartyRemote
 import io.kotest.assertions.throwables.shouldThrowExactly
+import io.kotest.core.spec.IsolationMode
 import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.matchers.result.shouldBeFailure
+import io.kotest.matchers.result.shouldBeSuccess
 import io.kotest.matchers.shouldBe
+import io.mockk.coEvery
+import io.mockk.coVerify
 import io.mockk.every
 import io.mockk.mockk
 import io.mockk.mockkStatic
 import io.mockk.unmockkAll
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.tasks.await
 
 class PartiesRepositoryTest : ShouldSpec(
 	{
+		extension(CoroutineTestExtension())
+		isolationMode = IsolationMode.InstancePerTest
 
 		val databaseReference: DatabaseReference = mockk()
-		val firebaseDatabase: FirebaseDatabase = mockk {
-			every { getReference("Parties") } returns databaseReference
-		}
-		val tested = PartiesRepository(firebaseDatabase)
+		val tested = PartiesRepository(
+			mockk<FirebaseDatabase> {
+				every { getReference("Parties") } returns databaseReference
+			},
+		)
+		val partyRemote: PartyRemote = mockk()
 
 		beforeSpec {
 			mockkStatic(Query::snapshots)
+			mockkStatic("kotlinx.coroutines.tasks.TasksKt")
 		}
 
 		afterSpec {
@@ -79,6 +91,43 @@ class PartiesRepositoryTest : ShouldSpec(
 			shouldThrowExactly<NoSuchElementException> {
 				tested.partyById("1").first()
 			}
+		}
+
+		should("add party with correct id") {
+			val newKey = "newKey"
+			val newPartyRemote: PartyRemote = mockk()
+			val childDatabaseReference: DatabaseReference = mockk {
+				every { key } returns newKey
+				coEvery { setValue(newPartyRemote).await() } returns mockk()
+			}
+			every { databaseReference.push() } returns childDatabaseReference
+			every { partyRemote.copy(newKey) } returns newPartyRemote
+
+			val result = tested.addParty(partyRemote)
+
+			coVerify {
+				childDatabaseReference.setValue(newPartyRemote).await()
+			}
+			result.shouldBeSuccess()
+		}
+
+		should("thrown an exception WHEN key is null") {
+			val childDatabaseReference: DatabaseReference = mockk {
+				every { key } returns null
+			}
+			every { databaseReference.push() } returns childDatabaseReference
+
+			val result = tested.addParty(partyRemote)
+
+			result.shouldBeFailure()
+		}
+
+		should("throw an exception when push fails") {
+			every { databaseReference.push() } throws Exception()
+
+			val result = tested.addParty(partyRemote)
+
+			result.shouldBeFailure()
 		}
 	},
 )


app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
```diff
@@ -20,6 +21,13 @@ class PartiesRepository @Inject constructor(
 			}
 		}
 
+	suspend fun addParty(partyRemote: PartyRemote) =
+		runCatching {
+			partiesReference.push().let {
+				it.setValue(partyRemote.copy(id = it.key ?: "")).await()
```
marcinWisnia: it.key can be null ? in which situation ? is it ok that in some situation we will add a party with empty id string ?

bartek977: in what case the key will be null? I'm worried empty string could break something elsewhere

app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
```diff
@@ -80,5 +94,33 @@ class PartiesRepositoryTest : ShouldSpec(
...
+			val result = tested.addParty(partyRemote)
+
+			coVerify {
+				childDatabaseReference.setValue(newPartyRemote).await()
+			}
+			result.shouldBeSuccess()
+		}
+
+		should("throw an exception when setValue fails") {
+			val errorMessage = "An exception"
```
marcinWisnia: I think the error message checking is not needed. I would go just with Exception(). Also You have only databaseReference.push() mocking for error, instead of databaseReference.push().setValue, according to Your test name :)

marcinWisnia (changes requested):

app/src/main/kotlin/com/intive/picover/parties/model/PartyRemote.kt
```diff
@@ -8,3 +8,9 @@ data class PartyRemote(
 	val title: String,
 	val description: String,
 )
+
+fun Party.toRemote() = PartyRemote(
+	id = id,
+	title = title,
+	description = description,
+)
```
MarekMacko: 1. Is there a particular reason to define it as an extension?
2. Think about dependency between these two classes. Is it a good place for such mapping? Maybe better would be to define it in `Party` file?

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
MarekMacko: That also leads to another conclusion, if it's acceptable to set an id to `""` then what it such case happen more than once? Is it correct to have two objects with the same `id`?

app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
MarekMacko: @kmaslowiec You should test only code, which is relevant for app business logic. In the production code you don't care about the error message, you don't have any logic connected with it so assertion on message is redundant. 
Of course it's ok for learning purposes but keep in mind that in bigger code base it makes your tests more verbose.

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
kmaslowiec: According to the documentation "
The last token in the location pointed to by this reference or null if this reference points to the database root". So very unlikely, but I get the point. 
Code updated.

kmaslowiec: According to the documentation "
The last token in the location pointed to by this reference or null if this reference points to the database root". So very unlikely, but I get the point. Code updated.

app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
kmaslowiec: Code updated.

app/src/main/kotlin/com/intive/picover/parties/model/PartyRemote.kt
kmaslowiec: @marcinWisnia What do you think, I think we have discussed it preiviously.

MarekMacko: @kmaslowiec  I'm asking about your opinion :)

kmaslowiec: IMO as it is mapping from Party to PartyRemote it would be better to stay with an approach that whichever, class is extended, there is the mapping. In other words, I agree :) But I believe, (I am not sure if I remember correctly) @marcinWisnia suggested otherwise, therefore, @marcinWisnia what do you think?

marcinWisnia: @kmaslowiec  @MarekMacko sorry, my mistake - i must have read something wrong - it should be in Party class as it is a "part" of this class functionality, that's my real feeling  :D.  

According to having it as an extension, i don't think it is really needed, but if someone has another opinion, please share :D

kmaslowiec: @marcinWisnia Ok. In that case what about this? 

`data class Party(
	val id: String,
	val title: String,
	val description: String,
)

fun List<PartyRemote>.toUI() =
	map { it.toUI() }

fun PartyRemote.toUI() =
	Party(
		id = id,
		title = title,
		description = description,
	)`

MarekMacko: Take into consideration potential coupling between these two classes. We have only one module but remote classes shouldn't know anything about UI. In the same way repository knowns nothing about activity or fragment, or at least shouldn't know :)

kmaslowiec: @MarekMacko I agree. I've moved the Party ext. to Party class and same with the RemoteParty.

MarekMacko: We've started discussion about placing extension `Party.toRemote()` in `PartyRemote.kt` file. Conclusion is to keep such extension in `Party.kt` file. 
Extension `PartyRemote.toUI` should stay as it was.

kmaslowiec: @MarekMacko Ok, I have re-read your comment regarding Remote Classes and UI and I think now I know what you mean. Sorry, I've just over-simplify it.

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
```diff
@@ -20,6 +21,15 @@ class PartiesRepository @Inject constructor(
 			}
 		}
 
+	suspend fun addParty(partyRemote: PartyRemote) =
+		runCatching {
+			partiesReference.push().let { databaseReference ->
+				databaseReference.key?.let { key ->
```
marcinWisnia, line 27: so what will happen now if the key is null ? just nothing will happen ? shouldn't we force the key here ?  is it possible to have no key here ?

kmaslowiec: @marcinWisnia According to the documentation "
The last token in the location pointed to by this reference or null if this reference points to the database root". So it is very unlikely.

MarekMacko: @kmaslowiec It's not an answer for @marcinWisnia questions :)

kmaslowiec: Ok, I will try to reproduce that tomorrow morning

MarekMacko: There is nothing to reproduce, just think about consequences :)

kmaslowiec: @MarekMacko @marcinWisnia If something happens with the Firebase Server, the part databaseReference.setValue(partyRemote.copy(id = key)).await() will not be called therefore, the Party will not be added. Unfortunately, nobody will know about it. I could generate a random number for this scenario but I would rather throw an Exception and later handle it with the SnackBar with an error message.

MarekMacko: @kmaslowiec Perfectly, that's the point and right conclusion :)

marcinWisnia: @kmaslowiec that's the spirit :D

kmaslowiec: @marcinWisnia @MarekMacko Updated. I was wondering if I should throw more specific error such as IllegalStateExcepption but I think a general error message 'Something went wrong" should be enough for such a corner case.

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
kmaslowiec: @marcinWisnia @MarekMacko Updated. I was wondering if I should throw more specific error such as IllegalStateExcepption but I think a general error message 'Something went wrong" should be enough for such a corner case.

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
```diff
@@ -15,3 +15,9 @@ fun PartyRemote.toUI() =
 		title = title,
 		description = description,
 	)
+
+fun Party.toRemote() = PartyRemote(
```
BorysChajdas: Why it is an extension? It could be written in a `Party` body, few lines above.

BorysChajdas (changes requested):

app/src/main/kotlin/com/intive/picover/parties/model/Party.kt
kmaslowiec: @BorysChajdas In  this PR I am focused on Repo. I will consider your suggestion within the next ViewModel PR. I am removing this for now to work on it later.

BorysChajdas: It's just about moving it a few lines up :D but okay

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
```diff
@@ -20,6 +21,15 @@ class PartiesRepository @Inject constructor(
...
+					databaseReference.setValue(partyRemote.copy(id = key)).await()
+				} ?: throw Exception()
```
BorysChajdas: I would vote for adding some error message, it could be very simple like `the key from Firebase is null`

app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
```diff
@@ -5,28 +5,41 @@ import com.google.firebase.database.DatabaseReference
...
 	{
+		extension(CoroutineTestExtension())
+		isolationMode = IsolationMode.InstancePerTest
 
 		val databaseReference: DatabaseReference = mockk()
 		val firebaseDatabase: FirebaseDatabase = mockk {
 			every { getReference("Parties") } returns databaseReference
 		}
 		val tested = PartiesRepository(firebaseDatabase)
+		val newKey = "newKey"
```
BorysChajdas: I don't think it's worth to store it as class field as it is only used in one test, same for other properties.

BorysChajdas (changes requested):

app/src/main/kotlin/com/intive/picover/parties/repository/PartiesRepository.kt
kmaslowiec: Good point, I hope the message is descriptive enough :)

kmaslowiec: @BorysChajdas updated

app/src/test/kotlin/com/intive/picover/parties/repository/PartiesRepositoryTest.kt
kmaslowiec: @BorysChajdas Fixed for all properties that occur only in one test.

BorysChajdas (approved):

marcinWisnia (approved):

PR #449, opened 2023-12-22
#448 Add warning to not update Kotlin before Compose Compiler

### LINKED ISSUE #448: Add warning to not update Kotlin before Compose Compiler
Image

kmaslowiec:
closes #448 

Please also see #447 

image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,3 +1,4 @@
+import com.android.build.gradle.internal.cxx.logging.warnln
 import com.google.firebase.appdistribution.gradle.firebaseAppDistribution
 import org.jetbrains.kotlin.noarg.gradle.NoArgExtension
 
@@ -44,6 +45,10 @@ android {
 
 	composeOptions {
 		kotlinCompilerExtensionVersion = "1.5.7"
+		warnln(
+			"The Compose Compiler 1.5.7 is not compatible with the newest version of Kotlin 1.9.22. " +
+				"Please update Kotlin once compatible Compose Compiler is available.",
+		)
 	}
 
 	kotlin {


BorysChajdas:
It only prints a warning or is it affecting renovate somehow? I'm wondering what we will get thanks to that?

app/build.gradle.kts
```diff
@@ -44,6 +45,10 @@ android {
 
 	composeOptions {
 		kotlinCompilerExtensionVersion = "1.5.7"
+		warnln(
+			"The Compose Compiler 1.5.7 is not compatible with the newest version of Kotlin 1.9.22. " +
```
MarekMacko, line 49: What is the point of this warning?

kmaslowiec: @MarekMacko The idea was to warn another dev to not update Kotlin to 1.9.22 as it would crash the app. But now I think it should be next to the place where the Kotlin version is listed. Just to save the time.

MarekMacko: there is a renovate bot cofigured. In this case it won't even build the project so you're not able to merge it.

kmaslowiec: @MarekMacko Yep, but it will take some time for somebody, most likely me, to figure it out again, and again when renovate will try to update the Kotlin. Then this comment can be useful. At this moment I am not sure if there was a clear the error of this, or I had to google it. But I guess the second.

BorysChajdas: I think everyone using Compose should remember that Compose needs to be updated to use a newer Kotlin version, so this warning is redundant to me. At first glance maybe it looks like it could be helpful but IMO it could also be misleading because the warning will need to be updated with the newer Kotlin version. Who will remember that? :D

kmaslowiec:
It has been discussed and the conclusion is that the comment is not necessary.

